PR #5, opened 2023-07-28
Fix a typo in the events page

Matyrobbrt:
The plural of `bus` is `buses`.

### DIFF
--- a/docs/concepts/events.md
+++ b/docs/concepts/events.md
@@ -7,7 +7,7 @@ Example: An event can be used to perform an action when a Vanilla stick is right
 
 The main event bus used for most events is located at `MinecraftForge#EVENT_BUS`. There is another event bus for mod specific events located at `FMLJavaModLoadingContext#getModEventBus` that you should only use in specific cases. More information about this bus can be found below.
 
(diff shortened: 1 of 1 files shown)


docs/concepts/events.md
```diff
@@ -7,7 +7,7 @@ Example: An event can be used to perform an action when a Vanilla stick is right
 
 The main event bus used for most events is located at `MinecraftForge#EVENT_BUS`. There is another event bus for mod specific events located at `FMLJavaModLoadingContext#getModEventBus` that you should only use in specific cases. More information about this bus can be found below.
 
-Every event is fired on one of these busses: most events are fired on the main forge event bus, but some are fired on the mod specific event buses.
+Every Forge event is fired on one of those buses: most events are fired on the main forge event bus, but some are fired on the mod specific event buses (generally registration events are fired on the mod bus).
```
sciwhiz12, line 10: Should we future-proof by not including the addition on registration events? As I recall, one of the items for discussion was moving most registration events away from the mod-specific event buses, as those buses are (to be) reserved for FancyModLoader/mod-loading events.

sciwhiz12 (changes requested):

PR #9, opened 2023-08-12
Migrate to Docusaurus frontmatter & change all heading levels to use markdown

Mysterious-Dev:
https://docusaurus.io/docs/markdown-features#front-matter

https://docusaurus.io/docs/markdown-features#standard-features

------------------
Preview URL: https://pr-9.neoforged-docs-previews.pages.dev


Mysterious-Dev:
Well, I think I finished the work.

For links, it will be better to do that in a separeted PR (for easy reviews), this PR is already big enough

ChampionAsh5357:
Why are we removing the titles and only sticking it in the metadata? We can just leave the title as is since it'll be consumed by the parser. The one I think should probably be in the metadata is the description if the first sentence doesn't accurately describe the page.

Mysterious-Dev:
Hey! Sorry for the delay for my reply. I did it out of reflex because I think it's more "cleaner" (as generally, front matters are used for page headers). But, if It's needed to change it, I can.

ChampionAsh5357:
All good. Though I think its better with the title since it stands out to me that I've clicked the correct page. Since we normally have  a quick description of each page below the header, it makes sense to have it there imo.

Mysterious-Dev:
So I should use h1 for the title?

ChampionAsh5357:
Yep, wrote up a whole style guide if you need it: https://docs.neoforged.net/contributing#style-guide

Mysterious-Dev:
@ChampionAsh5357 It's changed ^^

ChampionAsh5357 (changes requested):
There's not a file for this, but there likely needs to be a `_category_.json` file at the top level to order the folders.

Mysterious-Dev:
> There's not a file for this, but there likely needs to be a `_category_.json` file at the top level to order the folders.

Directly in docs folder ?

ChampionAsh5357:
We could also change from an autogenerated solution to modifying the sidebars.js so that everyone doesn't have to modify the position of every file.

Mysterious-Dev:
> We could also change from an autogenerated solution to modifying the sidebars.js so that everyone doesn't have to modify the position of every file.

I think it's the better solution ^^.

ChampionAsh5357:
Go for it. We'll probably need to change the contributing guidelines too to match that

ChampionAsh5357:
Go for it. We'll probably need to change the contributing guidelines too to match that

ChampionAsh5357:
Resolve the merge conflicts and this should be good to go. We'll see when new pages are added whether autogenerated subfolders make sense, or if we should do it manually.

ChampionAsh5357:
Closing as completed by #65.

PR #10, opened 2023-09-06
Add a detailed Mixin tutorial in advanced area.

Jissee:
I wrote a Mixin tutorial. I also warn the readers that they should use Events rather than Mixin for safety.

### DIFF
--- a/docs/advanced/mixin/3.inject.md
+++ b/docs/advanced/mixin/3.inject.md
@@ -1,0 +1,113 @@
+@Inject
+======
+Code injection is to "insert" your codes to a certain place of a certain vanilla method.
+
+Here, let's take ```ServerPlayer``` as an example.
+
+Basic
+-----
+
+### Determine Your Purpose
+
+In this example, you use Mixin to send the player a message when hurt.
+
+:::caution
+A more recommended approach is to listen to ```LivingHurtEvent``` and send the message on the event. Such an event can provide you with the specific player instance, while Mixin applies the codes to the class and all instances.
+:::
+
+### Find Target Methods
+
+You can find a method called ```hurt``` in ```ServerPlayer``` class, which takes two parameters: a ```DamageSource``` and a float indicates the amount.
+
+```java
+public boolean hurt(DamageSource p_9037_, float p_9038_){
+	...
+}
+```
+
+### Mixin Method
+
+Copy the method with the name and parameters to your mixin class. You can change the method and parameter names in the mixin class but do not change the parameter types and sequence.
+
+Add an additional parameter to this method, which depends on the return value. If the method is void then add ```CallbackInfo```. If the method returns a value then add ```CallbackInfoReturnable<ReturnType>```.
+
+Change the return type of the method to void.
+
+Add and complete the annotation ```@Inject``` Basically, it takes two parameters. ```method``` is the target method, and ```at``` is the injection position. The most common injection position is ```HEAD```, ```TAIL```, and ```RETURN```. The details can be found in the documents of Mixin.
+
+### Mixin Codes
+Then you can complete the codes in the mixin method, just like you are writing codes at the head of the vanilla method ```hurt```.
+
+However, the mixin method is in the mixin class and ```this``` refers to the mixin instance but not the player instance. So you can use a trick to get the "this" player instance. Cast ```this``` to ```Object``` and then ```ServerPlayer```. It may seem strange elsewhere, but it is normal in Mixin.
+
+Therefore, the complete method can be like this
+```java
+@Inject(method = {"hurt"}, at = {@At("HEAD")})
+public void onPlayerHurt(DamageSource source, float value, CallbackInfoReturnable<Boolean> ci){
+    ServerPlayer self = (ServerPlayer) (Object) this;
+    self.sendSystemMessage(Component.literal("Ouch!"));
+}
+
+```
+
+Save codes and run the game. You will receive the message every time you get hurt.
+
+:::tip
+To inject into non-parameterized constructors, or there is only one constructor no matter parameterized or non-parameterized, the method should be ```method = {"<init>"}```, but not ```method = {"init"}```.  
+
+To inject into parameterized ones, the method name will be discussed later.
+
+To inject into all constructors, the method should be ```method = {"<init>*"}```.
+
+The injection position of constructors can only be ```at = {@At("TAIL")}``` or ```at = {@At("RETRN")}```.
+
+To inject into static blocks, the method should be ```method = {"<clinit>"}```
+:::
+
+Advanced
+--------
+
+Sometimes there can be method overloading. Several methods with different parameter types have the same name. Such as ```sendSystemMessage(Component)``` and ```sendSystemMessage(Component, boolean)```
+
+In this situation, you should explicitly designate the method name and descriptor. Accordingly, the parameters of their mixin methods are also different.
+
+you should use this for the former one
+
+```java
+@Inject(method = {"sendSystemMessage(Lnet/minecraft/network/chat/Component;)V"}, at = {@At("HEAD")})
+public void onSendMessage(Component p_215097_, CallbackInfo ci){
+    ...
+}
+```
+
+and this for the latter one: 
+
+```java
+@Inject(method = {"sendSystemMessage(Lnet/minecraft/network/chat/Component;Z)V"}, at = {@At("HEAD")})
+public void onSendMessage(Component p_215097_, boolean p_240545_, CallbackInfo ci){
+    ...
+}
+```
+
+:::tip
+For overloaded constructors, the name is always ```method = {"<init>(...)V"}```. What you need to edit is the parameter type in (). 
+:::
+
+Here is a table of type descriptors, which is found in [Oracle's document][doc]. You can also find ways to complete method descriptors in that document.
+
+| FieldType term | Type      | Interpretation                                                                    |
+|----------------|-----------|-----------------------------------------------------------------------------------|  
+| B              | byte      | signed byte                                                                       |
+| C              | char      | Unicode character code point in the Basic Multilingual Plane, encoded with UTF-16 |
+| D              | double    | double-precision floating-point value                                             |
+| F              | float     | single-precision floating-point value                                             |
+| I              | int       | int	integer                                                                      |
+| J              | long      | long integer                                                                      |
+| L ClassName ;  | reference | an instance of class ClassName                                                    |
+| S              | short     | signed short                                                                      |
+| Z              | boolean   | true or false                                                                     |
+| [              | reference | one array dimension                                                               |
+
+
+
+[doc]: https://docs.oracle.com/javase/specs/jvms/se14/html/jvms-4.html#jvms-4.3.2
--- a/docs/advanced/mixin/4.shadow.md
+++ b/docs/advanced/mixin/4.shadow.md
@@ -1,0 +1,34 @@
+@Shadow
+======
+
+Sometimes you may want to access some private fields in the target class, but there is no getter. You may also want to access some private methods. If you are tired of [Access Transformer][at] and the obfuscation mapping, the annotation ```@Shadow``` can help you solve those problems.
+
+Fields
+------
+To access private fields in the target class, you just need to redefine the field with the same name in the mixin class and add the annotation ```@Shadow```. The shadowed fields should be private.
+
+```java
+@Shadow
+private <type name> <field name>;
+```
+
+Then you can use this field in your mixin methods just like normal fields. The Mixin system will automatically redirect your access to the corresponding fields in the target class.
+
+
+Methods
+-------
+To access private methods, you just need to redefine the method in the mixin class and add annotation ```@Shadow```. The shadowed methods should be protected abstract. Therefore, your mixin class must be abstract.
+
+```java
+@Shadow
+protected abstract <type name> <method name>(paramerer list...);
+```
+Then you can call this method in your mixin methods just like normal methods. The Mixin system will automatically redirect your access to the corresponding methods in the target class.
+
+:::tip
+Do not change field names, method names, parameter types, and sequence, or the shadow will not work.
+
+You can change parameter names.
+:::
+
+[at]: ../accesstransformers.md
--- a/docs/advanced/mixin/5.overwrite.md
+++ b/docs/advanced/mixin/5.overwrite.md
@@ -1,0 +1,13 @@
+@Overwrite
+=========
+:::caution
+This is ```@Overwrite``` but NOT ```@Override```.
+:::
+:::danger
+It can be dangerous to completely change vanilla mechanisms.
+:::
+Annotation ```@Overwrite``` allows you to discard all codes in a vanilla method and completely rewrite that method. It can be dangerous to completely change vanilla mechanisms.
+
+To overwrite a method, just create a method in your mixin class, whose name and parameter types and sequence should perfectly match the target method's. The only thing that can change is the parameter names. Then add the annotation ```@Overwrite``` to the method.
+
+
--- a/docs/advanced/mixin/index.md
+++ b/docs/advanced/mixin/index.md
@@ -1,0 +1,12 @@
+Mixin
+=====
+
+:::danger
+Before we start, we shall recommend you use [Events][events] instead of Mixin as it provides a safer way to inject your code. You should consider using Mixin if and only if you cannot find any satisfied Event and you must edit vanilla Minecraft codes.
+:::
+
+[SpongePowered Mixin][mixin_github] is a trait/mixin and bytecode weaving framework for Java using ASM. You can "edit" Minecraft vanilla codes and inject your own codes into Minecraft through Mixin.
+
+
+[mixin_github]: https://github.com/SpongePowered/Mixin
+[events]: ../../concepts/events.md
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,6 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+/.vs/Documentation/FileContentIndex
+/.vs/Documentation/v17
+/.vs
--- a/docs/advanced/mixin/1.preparation.md
+++ b/docs/advanced/mixin/1.preparation.md
@@ -1,0 +1,141 @@
+Preparation
+===========
+
+Edit Your ```build.gradle```
+----------------------
+
+Apply the mixin plugin.
+```groovy
+plugins {
+    id 'org.spongepowered.mixin' version '0.7.+'
+}
+```
+
+Add run arguments to gradle tasks.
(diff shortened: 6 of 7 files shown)


ChampionAsh5357 (changes requested):
Currently, this does not explain anything. Rather, it mentions how to copy-paste things with a brief explanation on some of the annotations. Documentation needs to explain what something is, when you should use it, why it's useful, and how to do so in a general manner. Examples can be included, but they are meant to be supplementary to the actual explanation.

Jissee:
Thanks for your recommendations. I have improved and changed most of the pages I uploaded. Please review them and contact me if there are any mistakes.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+Basic
+-----
+
+### Injection Method
+
+Injection methods are some special methods in Mixin that will not be invoked directly. Their codes will be injected into their target methods. Injection methods are annotated with ```@Inject```. Such an annotation is used to specify the target method, injection position, and other configurations.
+
+Two parameters are mandatory in  ```@Inject```:
+
+- ```method``` - The name of the target method, which should be in the target class. If the method is overloaded, it should be specified by the descriptors, which will be discussed later.
```
ChampionAsh5357, line 19: This can be multiple methods.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+- ```at``` - The injection position in the target method. 
+
+The most common positions for ```at``` are:
+
+- ```HEAD``` - At the start of the method
+- ```TAIL``` - At the end of the method
+- ```RETURN``` - Before every return statement of the method
+
+The name of the injection method can be different from the target method while the types and sequences of the parameters should not be changed. Additionally, there is an extra parameter that should be appended to the parameter list of the injection method.
```
ChampionAsh5357, line 28: Also half-true, the parameters can be different depending on the configurations within the annotation. Additionally, the callback parameter must be the last parameter: https://github.com/SpongePowered/Mixin/wiki/Advanced-Mixin-Usage---Callback-Injectors#2-building-handler-methods

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+    ...
+}
+```
+
+whose target is
+```java
+public boolean methodName(Type1 param1, Type2 param2){
+    ...
+}
+```
```
ChampionAsh5357, line 48: Where is it being injected into here?

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+Constructors are also treated as methods with name ```<init>``` but not ```init```. Here are some different situations when injecting into the constructors.
```
ChampionAsh5357, line 50: Explain why. Why is `<init>` the name of the constructor method?

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+- To inject into non-parameterized constructors, or there is only one constructor no matter whether parameterized or non-parameterized, the method should be ```method = {"<init>"}```.  
+
+- To inject into parameterized and overloaded constructors, the method should be specified by the descriptors, which will be discussed later.
```
ChampionAsh5357, line 54: Discuss this first. Explanations should not be pushed off later if its needed to understand how to use it.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+whose target is
+```java
+public boolean methodName(Type1 param1, Type2 param2){
+    ...
+}
+```
+
+Constructors are also treated as methods with name ```<init>``` but not ```init```. Here are some different situations when injecting into the constructors.
+
+- To inject into non-parameterized constructors, or there is only one constructor no matter whether parameterized or non-parameterized, the method should be ```method = {"<init>"}```.
```
ChampionAsh5357, line 52: Can also be `method = "<init>"`

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+- To inject into non-parameterized constructors, or there is only one constructor no matter whether parameterized or non-parameterized, the method should be ```method = {"<init>"}```.  
+
+- To inject into parameterized and overloaded constructors, the method should be specified by the descriptors, which will be discussed later.
+
+- To inject into all constructors, the method should be ```method = {"<init>*"}```.
+
+- The injection position of constructors can only be ```TAIL``` or ```RETURN```.
+
+- To inject into static blocks, the method should be ```method = {"<clinit>"}```
```
ChampionAsh5357, line 60: This can be applied to any method, not just constructors.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+Constructors are also treated as methods with name ```<init>``` but not ```init```. Here are some different situations when injecting into the constructors.
+
+- To inject into non-parameterized constructors, or there is only one constructor no matter whether parameterized or non-parameterized, the method should be ```method = {"<init>"}```.  
+
+- To inject into parameterized and overloaded constructors, the method should be specified by the descriptors, which will be discussed later.
+
+- To inject into all constructors, the method should be ```method = {"<init>*"}```.
+
+- The injection position of constructors can only be ```TAIL``` or ```RETURN```.
```
ChampionAsh5357, line 58: Again, explain for all cases and not just constructors.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+
+Constructors are also treated as methods with name ```<init>``` but not ```init```. Here are some different situations when injecting into the constructors.
+
+- To inject into non-parameterized constructors, or there is only one constructor no matter whether parameterized or non-parameterized, the method should be ```method = {"<init>"}```.  
+
+- To inject into parameterized and overloaded constructors, the method should be specified by the descriptors, which will be discussed later.
+
+- To inject into all constructors, the method should be ```method = {"<init>*"}```.
```
ChampionAsh5357, line 56: There needs to be a warning here for trying to consume more than one method.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+- The injection position of constructors can only be ```TAIL``` or ```RETURN```.
+
+- To inject into static blocks, the method should be ```method = {"<clinit>"}```
+
+
+
+### Injection Codes
+Injection codes are the codes in the injection methods that will be injected into the target methods.
```
ChampionAsh5357, line 65: I don't understand what this sentence means.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+Most time you can complete the injection codes as if they were in the corresponding places of target methods. But to access fields or invoke methods, you should use [Shadow][shadow].
```
ChampionAsh5357, line 67: When should I access fields or invoke methods?

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+- To inject into static blocks, the method should be ```method = {"<clinit>"}```
+
+
+
+### Injection Codes
+Injection codes are the codes in the injection methods that will be injected into the target methods.
+
+Most time you can complete the injection codes as if they were in the corresponding places of target methods. But to access fields or invoke methods, you should use [Shadow][shadow]. 
+
+Additionally,  ```this``` cannot be directly used to refer to the target instance as the codes are in the mixin class. You should cast ```this``` to ```Object``` and then to the target class. It may seem strange elsewhere, but it is normal in Mixin.
```
ChampionAsh5357, line 69: Explain why! It's not strange elsewhere, it's bad practice almost anywhere. It's done for a very specific reason.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+```java
+TargetClass self = (TargetClass) (Object) this;
+```
+
+If you want to end the methods in advance, or "return" the target method in your injection codes, you should add an optional parameter ```cancellable``` to the annotation.
+
+```java
+@Inject(method = {""}, at = {@At("")}, cancellable = true)
```
ChampionAsh5357, line 78: Give a more complete example.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+| B              | byte      | signed byte                                                                       |
+| C              | char      | Unicode character code point in the Basic Multilingual Plane, encoded with UTF-16 |
+| D              | double    | double-precision floating-point value                                             |
+| F              | float     | single-precision floating-point value                                             |
+| I              | int       | int	integer                                                                      |
+| J              | long      | long integer                                                                      |
+| L ClassName ;  | reference | an instance of class ClassName                                                    |
+| S              | short     | signed short                                                                      |
+| Z              | boolean   | true or false                                                                     |
+| [              | reference | one array dimension                                                               |
```
ChampionAsh5357, line 104: This is already in the AT docs correct? This should probably be moved into its own section then.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+To inject into overloaded methods, you should explicitly designate the method name as well as their descriptors in the ```()```. The following [Oracle's document][doc] can help you to finish this. Accordingly, the parameters of their injection methods are also different.
+
+| FieldType term | Type      | Interpretation                                                                    |
+|----------------|-----------|-----------------------------------------------------------------------------------|  
+| B              | byte      | signed byte                                                                       |
+| C              | char      | Unicode character code point in the Basic Multilingual Plane, encoded with UTF-16 |
+| D              | double    | double-precision floating-point value                                             |
+| F              | float     | single-precision floating-point value                                             |
+| I              | int       | int	integer                                                                      |
```
ChampionAsh5357, line 99: Table formatting.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+}
+```
+and
+
+```java
+@Inject(method = {"messageName(II)V"}, at = {@At("HEAD")})
+public void onMethodName(int param1, int param2, CallbackInfo ci){
+    ...
+}
+```
```
ChampionAsh5357, line 121: Show target example.

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+:::tip
+For overloaded constructors, the name is always ```method = {"<init>(...)V"}```. What you need to edit is the parameter type in (). 
+:::
```
ChampionAsh5357, line 125: Example!

docs/advanced/mixin/3.inject.md
```diff
@@ -0,0 +1,133 @@
...
+
+
+
+
+
```
ChampionAsh5357, line 130: Empty space

docs/advanced/mixin/4.shadow.md
```diff
@@ -0,0 +1,46 @@
+@Shadow
+======
+
+Shadow is a technology that redirects the access of the fields and methods from the mixin classes to the vanilla classes. It allows you to use those elements like normal ones so that your method code can be simplified. It can also help you to access the private fields or methods in the target class.
```
ChampionAsh5357, line 4: This also needs a reword. Shadow acts like a placeholder for the actual field or method call. I wouldn't call it necessarily a redirect because of the bytecode merging replaces the calls.

docs/advanced/mixin/4.shadow.md
```diff
@@ -0,0 +1,46 @@
...
+Shadow is a technology that redirects the access of the fields and methods from the mixin classes to the vanilla classes. It allows you to use those elements like normal ones so that your method code can be simplified. It can also help you to access the private fields or methods in the target class. 
+
+Fields
+------
+Shadowed fields are annotated with  ```@Shadow``` and have the same access modifiers, names, and types as the vanilla fields in the target class. 
+
+```java
+@Shadow
+<access modifier> <static or not> <type name> <field name>;
+```
```
ChampionAsh5357, line 13: Example is better here since everyone should know how to create a field in a class.

docs/advanced/mixin/4.shadow.md
```diff
@@ -0,0 +1,46 @@
...
+
+The Mixin system will automatically redirect your access to the corresponding fields in the target class.
```
ChampionAsh5357, line 15: Not really true since it depends on other compatibilities. Edge cases should be documented.

docs/advanced/mixin/4.shadow.md
```diff
@@ -0,0 +1,46 @@
...
+-------
+Shadowed methods are annotated with  ```@Shadow``` and have the same access modifiers, names, parameter types, and return types as the vanilla methods in the target class.
+
+If the vanilla method is not static, the shadowed method should be abstract.
+
+
+```java
+@Shadow
+<access modifier> abstract <type name> <method name>(parameter list...);
+```
```
ChampionAsh5357, line 28: Example than abstractness here.

docs/advanced/mixin/4.shadow.md
```diff
@@ -0,0 +1,46 @@
...
+
+If the vanilla method is static, the shadowed method should be static and have a method body. The return value does not matter because this method is just a placeholder.
+```java
+@Shadow
+<access modifier> static <type name> <method name>(parameter list...){
+	return null;       // for reference types
+	// return 0;       // for primitives
+	// return false;
+}
+```
```
ChampionAsh5357, line 38: I believe most people throw an exception.

docs/advanced/mixin/4.shadow.md
```diff
@@ -0,0 +1,46 @@
...
+
+The Mixin system will automatically redirect your access to the corresponding methods in the target class.
```
ChampionAsh5357, line 40: Not really true since it depends on other compatibilities. Edge cases should be documented.

docs/advanced/mixin/4.shadow.md
```diff
@@ -0,0 +1,46 @@
...
+
+:::tip
+Do not change field names, method names, parameter types, and sequences, or the shadow will not work.
+
+You can change parameter names.
+:::
```
ChampionAsh5357, line 46: Half-true, see `Shadow#prefix`: https://github.com/SpongePowered/Mixin/blob/155314e6e91465dad727e621a569906a410cd6f4/src/main/java/org/spongepowered/asm/mixin/Shadow.java#L87

docs/advanced/mixin/5.overwrite.md
```diff
@@ -0,0 +1,53 @@
+@Overwrite
+=========
+:::caution
+This is ```@Overwrite``` but NOT ```@Override```.
+:::
```
ChampionAsh5357, line 5: Why is this a caution?

docs/advanced/mixin/5.overwrite.md
```diff
@@ -0,0 +1,53 @@
...
+
+Overwriting is a technology that discards all codes in a vanilla method and replaces it with your own code. It is a thorough way to change the vanilla mechanisms.
+
+:::danger
+It can be dangerous to completely change vanilla mechanisms.
+:::
```
ChampionAsh5357, line 11: Why? This really needs to be explained better.

docs/advanced/mixin/5.overwrite.md
```diff
@@ -0,0 +1,53 @@
...
+
+You should not consider overwriting unless:
+
+- You want to completely change vanilla mechanisms.
+- You want to remove some codes from these vanilla methods.
+- You want to fix the bugs in these vanilla methods.
+- You have a better implementation than these vanilla codes.
```
ChampionAsh5357, line 18: None of these are potentially good reasons. Overwrites should be used if there is no other potential alternative. You can do most of these things with injections in 90% of cases.

docs/advanced/mixin/5.overwrite.md
```diff
@@ -0,0 +1,53 @@
...
+
+You can write overwriting methods in your mixin class, which will be applied to the target class.
+
+Overwriting methods are annotated with ```@Overwrite``` and they must have the same names, parameter types, and sequences as the vanilla methods. What can be changed is only the parameter names. Otherwise, the overwriting will not take effect.
```
ChampionAsh5357, line 22: This should probably explain about param names not really meaning anything in bytecode land with the LVT.

docs/advanced/mixin/5.overwrite.md
```diff
@@ -0,0 +1,53 @@
...
+
+Once you have overwritten some method, you should add the JavaDoc, including the author's name and reasons, to the method.
+```java
+/**
+* @author
+* @reason
+*/
+@Overwrite
+...
+```
```
ChampionAsh5357, line 49: `@reason` is not a valid javadoc tag by default. Also, you should always document your code regardless.

docs/advanced/mixin/5.overwrite.md
```diff
@@ -0,0 +1,53 @@
...
+
+:::tip
+You cannot overwrite the constructors.
+:::
```
ChampionAsh5357, line 53: Explain why.

docs/advanced/mixin/index.md
```diff
@@ -0,0 +1,12 @@
+Mixin
+=====
+
+:::danger
+Before we start, we shall recommend you use [Events][events] instead of Mixin as it provides a safer way to inject your code. You should consider using Mixin if and only if you cannot find any satisfied Event and you must edit vanilla Minecraft codes.
+:::
+
+[SpongePowered Mixin][mixin_github] is a trait/mixin and bytecode weaving framework for Java using ASM. You can "edit" Minecraft vanilla codes or inject your own codes into Minecraft through Mixin.
```
ChampionAsh5357, line 8: This needs a better introduction on what this is.

ChampionAsh5357 (changes requested):
More explanations are required. Also, if one thing is documented, all or at least the most common usecases should be documented as well. There are numerous things that are glossed over here that have an important reason why they were designed that way.

Jissee:
Thanks for your concern and recommendations, which are really helpful for me. I am only a college student and a beginner at Minecraft modding actually. I wrote those documents based on my development experience. Sorry for my incomplete document contribution. I am a little tired these days and I am afraid that I cannot answer some of your questions. You can close this PR if this bothers you and I will continue to study in the future. Again, thanks for your patience.

ChampionAsh5357:
It's fine. If you don't have the time to contribute or address the issues, then you're welcome to leave it alone or close at a later point in time depending on if you ever wish to come back to this.

PR #14, opened 2023-10-02
Contributing Guidelines for the Documentation Repository

ChampionAsh5357:
This defines the guidelines for contributing to this repo along with the style guide for markdown files. Additionally, this adds a `.node-version` file to use nvm and nvs with the correct version of Node.

------------------
Preview URL: https://pr-14.neoforged-docs-previews.pages.dev

PR #15, opened 2023-10-03
create attribute documentation

justliliandev:
This adds documentation for the vanilla attribute system

------------------
Preview URL: https://pr-15.neoforged-docs-previews.pages.dev


docs/entities/_category_.json
```diff
@@ -0,0 +1,3 @@
+{
+    "label": "Entities"
+}
```
ChampionAsh5357, line 3: New line at EOF.

docs/entities/attributes.md
```diff
@@ -0,0 +1,59 @@
+Attributes
+==========
+
+Attributes are ways to modify specific values for living entities in a mod compatible way.
```
ChampionAsh5357, line 4: That's not really what an attribute is. That's what we use it for, yes, but it's not the intended purpose. Iirc, attributes are a property defined on an entity that is meant to apply the greatest modifier without stacking to the player.

Please correct me if I'm wrong, I can't look on my computer because it's a hot brick.

docs/entities/attributes.md
```diff
@@ -0,0 +1,59 @@
...
+| `FLYING_SPEED`                | The flying speed of the entity. Not applicable to players who fly with the elytra            |
+| `ATTACK_DAMAGE`               | how much damage the entity deals when attacking                                              |
+| `ATTACK_KNOCKBACK`            | how much knockback the entity causes when physically attacking                               |
+| `ATTACK_SPEED`                | how many full damage attack the entity deals per second                                      |
+| `ARMOR`                       | how many armor points the entity has and how resistant it is to attacks                      |
+| `ARMOR_TOUGHNESS`             | how resistant the entity is to strong attacks                                                |
+| `LUCK`                        | changes the chance for treasure, junk and normal fishing loottables                          |
+| `SPAWN_REINFORCEMENTS_CHANCE` | how likely a zombie spawns reinforcements when attacked                                      |
+| `JUMP_STRENGTH`               | how high the entity jumps. Only applicable for horse like more (including mules and camels). |
```
ChampionAsh5357, line 24: I have no idea how useful this is. This makes sense in parchment, but I feel like this could also just be said it's in the attributes file and list some examples.

docs/entities/attributes.md
```diff
@@ -0,0 +1,59 @@
...
+NeoForge adds the following attributes:
+
+| Attribute              | Description                                                            |
+|------------------------|------------------------------------------------------------------------|
+| `SWIM_SPEED`           | how fast the entity swims                                              |
+| `NAMETAG_DISTANCE`     | how far the nametag can be seen for this entity                        |
+| `ENTITY_GRAVITY`       | how fast the entity accelerates when not on ground                     |
+| `BLOCK_REACH`          | how far the player can interact with the world                         |
+| `ENTITY_REACH`         | how far the player can interact with entities                          |
+| `STEP_HEIGHT_ADDITION` | how high the entity can walk up blocks in addition to Entity#maxUpStep |
```
ChampionAsh5357, line 35: Again, same issue here. This one makes more sense though to have in the documentation.

docs/entities/attributes.md
```diff
@@ -0,0 +1,59 @@
...
+
+Attribute, AttributeInstance and AttributeModifier
+-----------
+
+Attributes have to be [registered][registration] and are used to look up the current `AttributeInstance` for a given LivingEntity. The `AttributeInstance` holds all `AttributeModifier` currently active on the entity.
+AttributeModifiers require a UUID that should be unique to this modifier. It's used to ensure that an `AttributeModifier` can only be applied once to a LivingEntity. You should generate a random one once and then hardcode it. They also require an Operation and an amount to perform the calculation. How the calculation works can be read in the next section.
+The value of an `Attribute` for a given `LivingEntity` can be obtained by the `AttributeMap` for the LivingEntity and getting the `AttributeInstance` from it and then calling `AttributeInstance#getValue`.
```
ChampionAsh5357, line 42: Probably best to give an example on how to do so.

docs/entities/attributes.md
```diff
@@ -0,0 +1,59 @@
...
+| `ENTITY_GRAVITY`       | how fast the entity accelerates when not on ground                     |
+| `BLOCK_REACH`          | how far the player can interact with the world                         |
+| `ENTITY_REACH`         | how far the player can interact with entities                          |
+| `STEP_HEIGHT_ADDITION` | how high the entity can walk up blocks in addition to Entity#maxUpStep |
+
+Attribute, AttributeInstance and AttributeModifier
+-----------
+
+Attributes have to be [registered][registration] and are used to look up the current `AttributeInstance` for a given LivingEntity. The `AttributeInstance` holds all `AttributeModifier` currently active on the entity.
+AttributeModifiers require a UUID that should be unique to this modifier. It's used to ensure that an `AttributeModifier` can only be applied once to a LivingEntity. You should generate a random one once and then hardcode it. They also require an Operation and an amount to perform the calculation. How the calculation works can be read in the next section.
```
ChampionAsh5357, line 41: Why should I hardcode the UUID?

ChampionAsh5357, line 41: `Operation` can be explained in a subcategory here.

docs/entities/attributes.md
ChampionAsh5357, line 42: Example. Also, I don't think this is as useful as explaining what methods in the entity provide this.

docs/entities/attributes.md
```diff
@@ -0,0 +1,59 @@
...
+The value of an `Attribute` for a given `LivingEntity` can be obtained by the `AttributeMap` for the LivingEntity and getting the `AttributeInstance` from it and then calling `AttributeInstance#getValue`.
+
+Operations
+-----------
+
+AttributeInstances can have 3 different operations, `ADDITION`, `MULTIPLY_BASE` and `MULTIPLY_TOTAL`.
+First the default value is used and all `ADDITION`-Modifiers are added to it. This is the base value.
+`MULTIPLY_BASE`-Modifiers are applied next, they multiply the base value with 1 + the sum of the `MULTIPLY_BASE` modifiers.
+`MULTIPLY_TOTAL`-Modifiers are applied last, they multiply the value created with 1+modifier each.
```
ChampionAsh5357, line 50: Show maths.

ChampionAsh5357 (changes requested):
Take a look at #14 for the current draft standards.

In general, I think this needs more work on explaining what an attribute is, why do we use it, and how we do so.

docs/entities/attributes.md
agnor99: nope, it's not about the greatest modifier, all active AttributeModifiers are considered. addition stacks by just addition, multiply_base stacks by adding the factors (+20% and +20% combine to +40%) and multiply_total stacks by multiplication (+20% and +20% combine to +44%)

docs/entities/attributes.md
agnor99: I do believe that the list of attributes and what they do is useful. There are already parchment comments for them, but giving people examples for what mojang and neo use attributes for, so people can easier understand, what things attributes can be used for and therefore be used for their use cases

docs/entities/attributes.md
ChampionAsh5357: Okay. So the `AttributeMap` has a map of `Attribute`s to `AttributeInstance`s, which in turn hold a set of `AttributeModifier`s, unique by their id, grouped by operation.

Okay then, then it should still be something like: attributes are a property defined on an entity that can be modified by many different elements at the same time. Each element can be applied and removed depending on their intent.

There's probably a better way to word that, but it's what I'm thinking rn.

Shadows-of-Fire: An apt description is "Attributes are properties of an entity which correspond to a value. This value can be modified by multiple sources at once through Attribute Modifiers, allowing for simple cross-mod compatibility."

docs/entities/attributes.md
Shadows-of-Fire: This is one of those instances where I'm inclined to agree that the docs should delegate to the code for listing the attributes.  This is almost guaranteed to go stale, while a reference to the `Attributes` class will always be accurate.

docs/entities/attributes.md
Shadows-of-Fire: This is overly verbose IMO.  The quick method to retrieve the value of an attribute is `LivingEntity#getAttributeValue(Attribute)`

docs/entities/attributes.md
Shadows-of-Fire: The question of "should I hardcode the UUID" is implementation dependent on the answer of what will you need to do with that modifier in the future.

If you are performing a one-off permanent modifier addition (to say, an entity when it is being spawned), there is no need to use a static UUID.
If you are adding attribute modifiers to an item, you _must_ use a static UUID otherwise the modifier will not be removed when the item is unequipped.

Attributes also only permit one modifier with the same UUID (though one UUID can be used across Attributes), so it can be useful to prevent re-application.

Shadows-of-Fire: I have an apt description of Operation here: <https://github.com/Shadows-of-Fire/Apothic-Attributes/blob/1.20/src/main/java/dev/shadowsoffire/attributeslib/api/AttributeHelper.java#L52-L63>

Shadows-of-Fire:
Fair amount of work probably needed here.

The current state is unfortunately inbetween "too technical" and "too simple" in that it is not technical enough for advanced users to gain anything that they would not have inferred on their own, but also too verbose for new users.

IchHabeHunger54:
Will be superseded by #171 . Also has been inactive for a year now.

PR #17, opened 2023-10-04
Fix markdown link in components.md

tmvkrpxl0:
😅 I made a huge mistake when writing this, confused between internalization and internationalization

------------------
Preview URL: https://pr-17.neoforged-docs-previews.pages.dev


tmvkrpxl0:
https://github.com/neoforged/Documentation/blob/main/docs/misc/components.md?plain=1#L10
This uses [internationalization] so this one is definitely broken now

ChampionAsh5357 (approved):
Sure, it doesn't really make a difference either way, but it's good to have the correct word.

Issue #20, opened 2023-10-11
Update the Documentation to Match Contributing Guidelines


ChampionAsh5357:
The contributing guidelines have recently been established, so every page needs to be updated to support the new guidelines. To that end, every page will be reviewed and rewritten/reformatted as required. The changes will target the 20.2 version of NeoForge when released.

I have assigned myself since I will eventually redo all the pages, but anyone is welcome to help contribute.

- [ ] Getting Started with Neo
    - [ ] Mod Files
    - [ ] Structuring Your Mod
    - [ ] Versioning
- [ ] Advanced Topics
    - [ ] Access Transformers
- [ ] Block Entities
    - [ ] BlockEntityRenderer
- [ ] Blocks
    - [ ] Block States
- [ ] Concepts
    - [ ] Events
    - [ ] Internationalization and Localization
    - [ ] Mod Lifecycle
    - [ ] Registries
    - [ ] Resources
    - [ ] Sides in Minecraft
- [ ] Data Generators
    - [ ] Client
        - [ ] Language Generation
        - [ ] Model Generation
        - [ ] Sound Definition Generation
    - [ ] Server
        - [ ] Advancement Generation
        - [ ] Datapack Registry Object Generation
        - [ ] Global Loot Modifier Generation
        - [ ] Loot Table Generation
        - [ ] Recipe Generation
        - [ ] Tag Generation
- [ ] Data Storage
    - [ ] The Capability System
    - [ ] Codecs
    - [ ] Saved Data
- [ ] Game Effects
    - [ ] Particles
    - [ ] Sounds
- [ ] GUIs
    - [ ] Menus
    - [ ] Screens
- [ ] Items
    - [ ] BlockEntityWithoutLevelRenderer
- [ ] Legacy
    - [ ] Porting to Minecraft 1.20
- [ ] Miscellaneous
    - [ ] Text Components
    - [ ] Configuration
    - [ ] Debug Profiler
    - [ ] Game Tests
    - [ ] Key Mappings
    - [ ] Forge Update Checker
- [ ] Networking
    - [ ] Entities
    - [ ] SimpleImpl
- [ ] Rendering
    - [ ] Model Extensions
        - [ ] Face Data
        - [ ] Render Types
        - [ ] Root Transforms
        - [ ] Part Visibility
    - [ ] Custom Model Loaders
        - [ ] BakedModel
        - [ ] ItemOverrides
        - [ ] Transform
- [ ] Resources
    - [ ] Resource Packs
[...]

PR #21, opened 2023-10-17
Explain blocks and blockstates in more depth

IchHabeHunger54:
This PR improves and expands on the explanation of blocks and blockstates. More specifically, it:

- Adds more information on how every block and every blockstates exists in memory once, and why calling `new Block()` is a bad idea
- Adds more information on when to use blocks, when to use blockstates, and when to use block entities
- Adds a brief explanation of blockstate JSONs and block model JSONs, with references to the Resources section
- Adds information about how most methods expect a `BlockState` rather than a `Block`, and why that is the case
- Adds more information about `BlockState`'s most important usecases, specifically `BlockState#getProperty`, `BlockState#setProperty`, `Level#getBlockState` and `Level#setBlock`
- Adds information on the update flags for `Level#setBlock`
- Removes the historical section about block metadata, as it is irrelevant to the average modder
- Removes the disclaimer about optional registration, as that is better suited for the registration section
- Rewords some sections and fixes some grammar and spelling issues

------------------
Preview URL: https://pr-21.neoforged-docs-previews.pages.dev


docs/blocks/index.md
```diff
@@ -1,51 +1,61 @@
...
 
 :::note
-Blocks have no setter for their `CreativeModeTab`. This is handled by the [`BuildCreativeModeTabContentsEvent`][creativetabs] if the block has an associated item (e.g. `BlockItem`). Furthermore, there is no setter for translation key of the block as it is generated from the registry name via `Block#getDescriptionId`.
+It is important to understand that a block in the world is not the same thing as in an inventory. What looks like a block in an inventory is actually a `BlockItem`, a special type of item that places a block when used. This also means that things like the creative tab or the max stack size are handled by the corresponding `BlockItem`.
+
+A `BlockItem` must be registered separately from the block, this has the reason that not every block is intended to have a corresponding item.
 :::
 
-### Advanced Blocks
+### More functionality
```
ChampionAsh5357, line 35: Capitalization.

docs/blocks/index.md
```diff
@@ -1,51 +1,61 @@
...
+If you register your block and place it in the world, you will find it to be missing things like a texture. This is because textures, among others, are handled by Minecraft's resource system.
 
-When a block is registered, *only* a block is registered. The block does not automatically have an `BlockItem`. To create a basic `BlockItem` for a block, one should set the registry name of the `BlockItem` to that of its `Block`. Custom subclasses of `BlockItem` may be used as well. Once an `BlockItem` has been registered for a block, `Block#asItem` can be used to retrieve it. `Block#asItem` will return `Items#AIR` if there is no `BlockItem` for the `Block`, so if you are not certain that there is an `BlockItem` for the `Block` you are using, check for if `Block#asItem` returns `Items#AIR`.
-:::
+To apply a simple texture to a block, you must add a blockstate JSON, a model JSON, and a texture. See the section on [resources] for more information.
```
ChampionAsh5357, line 45: Probably should link to the actual resource page or MC wiki links. This doesn't tell me much.

docs/blocks/index.md
```diff
@@ -1,51 +1,61 @@
...
+Most places where you'd expect a block actually take a `BlockState` instead. For example, `Level#setBlock` takes a `BlockState`, not a `Block`, despite the misleading name. If you have a block that does not have any blockstate properties, you just have one blockstate on that block. That one blockstate can be retrieved by calling `Block#defaultBlockState`.
 
-Further Reading
----------------
+Similarly, most methods in `Block` that can be overridden by you have a `BlockState` as one of their parameters. This allows you to perform different behavior for different blockstates. For example, if a door is right-clicked, what happens depends on whether the door is currently open or closed.
```
ChampionAsh5357, line 52: Probably should state almost all here. Example would be nice too.

docs/blocks/states.md
```diff
@@ -1,54 +1,32 @@
...
+The `BlockState` system is a flexible and powerful system, but it also has limitations. `BlockState`s are immutable, and all combinations of their properties are generated on startup of the game. This means that having a `BlockState` with many properties and possible values will slow down the loading of the game, take up extra RAM, and may end up making your code confusing.
 
-The previous system of meaningless metadata values were replaced by a system of block properties, which are easier to interpret and deal with. Previously, a stone button which is facing east and is powered or held down is represented by "`minecraft:stone_button` with metadata `9`. Now, this is represented by "`minecraft:stone_button[facing=east,powered=true]`".
+Immutability has its upsides as well, though. For example, `==` can and should be used when comparing `BlockState`s. Another upside is that, using `BlockState#setValue(Property<?>)`, you can very easily go from one state to another. This does not actually create a new `BlockState`; it returns the previously generated state with that value instead.
```
ChampionAsh5357, line 15: This is a weird pros and cons list which doesn't really help me understand the benefits of `BlockState` compared to something like `BlockEntity`.

docs/blocks/states.md
```diff
@@ -1,54 +1,32 @@
...
 
-Proper Usage of Block States
----------------------------------------
+`BlockState` is also a final class, meaning it cannot be extended. **Any functionality goes in the corresponding Block class!**
```
ChampionAsh5357, line 17: Link!

docs/blocks/states.md
```diff
@@ -1,54 +1,32 @@
...
 
-The `BlockState` system is a flexible and powerful system, but it also has limitations. `BlockState`s are immutable, and all combinations of their properties are generated on startup of the game. This means that having a `BlockState` with many properties and possible values will slow down the loading of the game, and befuddle anyone trying to make sense of your block logic.
+When to use blockstates
+-----------------------
 
-Not all blocks and situations require the usage of `BlockState`; only the most basic properties of a block should be put into a `BlockState`, and any other situation is better off with having a `BlockEntity` or being a separate `Block`. Always consider if you actually need to use blockstates for your purposes.
+Only the most basic properties of a block should be put into a `BlockState`, sometimes using a `BlockEntity` or multiple separate `Block`s is the better approach. Always consider if you actually need to use blockstates for your purposes.
```
ChampionAsh5357, line 22: How does this help me understand when to use each?

docs/blocks/states.md
```diff
@@ -1,54 +1,32 @@
...
 
-:::note
-A good rule of thumb is: **if it has a different name, it should be a separate block**.
-:::
+A good rule of thumb for that is: **if it has a different name, it should be a separate block**.  An example is making chair blocks: the direction of the chair should be a property, while the different types of wood should be separated into different blocks. So you'd have one chair block for each wood type, and each chair block has four blockstates (one for each rotation).
```
ChampionAsh5357, line 24: Where did the `BlockEntity` go?

docs/blocks/states.md
```diff
@@ -1,54 +1,32 @@
...
 
-An example is making chair blocks: the *direction* of the chair should be a *property*, while the different *types of wood* should be separated into different blocks.
-An "Oak Chair" facing east (`oak_chair[facing=east]`) is different from a "Spruce Chair" facing west (`spruce_chair[facing=west]`).
+Implementing blockstates
```
ChampionAsh5357, line 26: Capitalization Part 4 (probably more but I haven't been counting)

docs/blocks/states.md
```diff
@@ -1,54 +1,32 @@
...
+------------------------
 
-Implementing Block States
----------------------------------------
-
-In your Block class, create or reference `static final` `Property<?>` objects for every property that your Block has. You are free to make your own `Property<?>` implementations, but the means to do that are not covered in this article. The vanilla code provides several convenience implementations:
+In your block class, create or reference `static final Property<?>` constants for every property that your block has. You are free to make your own `Property<?>` implementations, but the means to do that are not covered in this article. The vanilla code provides several convenience implementations:
```
ChampionAsh5357, line 29: Why should I do this? And why wouldn't it be covered?

docs/blocks/states.md
```diff
@@ -66,13 +44,12 @@ In your Block class, create or reference `static final` `Property<?>` objects fo
 
 The class `BlockStateProperties` contains shared vanilla properties which should be used or referenced whenever possible, in place of creating your own properties.
 
-When you have your desired `Property<>` objects, override `Block#createBlockStateDefinition(StateDefinition$Builder)` in your Block class. In that method, call `StateDefinition$Builder#add(...);`  with the parameters as every `Property<?>` you wish the block to have.
+When you have your desired property constants, override `Block#createBlockStateDefinition(StateDefinition$Builder)` in your block class. In that method, call `StateDefinition$Builder#add(...);`  with the parameters being all the properties you wish the block to have - typically this would be all the properties you defined before.
```
ChampionAsh5357, line 47: Example!

docs/blocks/states.md
```diff
@@ -66,13 +44,12 @@ In your Block class, create or reference `static final` `Property<?>` objects fo
...
 
-Every block will also have a "default" state that is automatically chosen for you. You can change this "default" state by calling the `Block#registerDefaultState(BlockState)` method from your constructor. When your block is placed it will become this "default" state. An example from `DoorBlock`:
+Every block will also have a default state. Usually, the default state uses the default value of every property. You can change the default state by calling the `Block#registerDefaultState(BlockState)` method from your constructor. For example, this is what the method call looks like in `DoorBlock`:
 
 ```java
-this.registerDefaultState(
-  this.stateDefinition.any()
+this.registerDefaultState(this.stateDefinition.any()
```
ChampionAsh5357, line 52: Why?

docs/blocks/states.md
```diff
@@ -66,13 +44,12 @@ In your Block class, create or reference `static final` `Property<?>` objects fo
...
+When you have your desired property constants, override `Block#createBlockStateDefinition(StateDefinition$Builder)` in your block class. In that method, call `StateDefinition$Builder#add(...);`  with the parameters being all the properties you wish the block to have - typically this would be all the properties you defined before.
 
-Every block will also have a "default" state that is automatically chosen for you. You can change this "default" state by calling the `Block#registerDefaultState(BlockState)` method from your constructor. When your block is placed it will become this "default" state. An example from `DoorBlock`:
+Every block will also have a default state. Usually, the default state uses the default value of every property. You can change the default state by calling the `Block#registerDefaultState(BlockState)` method from your constructor. For example, this is what the method call looks like in `DoorBlock`:
```
ChampionAsh5357, line 49: Not usually, it does use the default value for every property unless specified otherwise.

ChampionAsh5357 (changes requested):
More explanations and examples are necessary. Also, if you are rewriting the page, use the notes in the new contribution guidelines.

ChampionAsh5357:
Yeah, generally prefer pseudocode like java with comments over exact examples unless necessary. The guidelines should show an example

IchHabeHunger54:
Rewrote major parts of it to consider the feedback, examples have been added. I also added a WIP method reference page that I will finish tomorrow. Converting to draft for now.

docs/blocks/index.md
```diff
@@ -1,51 +1,61 @@
...
+Creating blocks
+---------------
+
+### Basic blocks
+
+For simple blocks which need no special functionality (think cobblestone, wooden planks, etc.), the `Block` class can be used directly. To do so, during registration, instantiate `Block` with a `BlockBehaviour$Properties` parameter. This `BlockBehaviour$Properties` parameter can be created using `BlockBehaviour$Properties#of`, and it can be customized by calling its methods. For instance:
+
+- `strength` - This method takes two parameters: the hardness and the resistance.
+    - The hardness is an arbitrary value that controls the time it takes to break the block. For reference, stone has a hardness of 1.5, dirt has a hardness of 0.5, obsidian has a hardness of 50, and unbreakable blocks (such as bedrock) have a hardness of -1.
```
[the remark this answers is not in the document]
IchHabeHunger54: As far as I'm aware, there's no known value to time conversion rate for this. I moved this to the method reference, but I'm unsure what to do about this.

docs/blocks/index.md
```diff
@@ -1,51 +1,61 @@
...
+    - The resistance is an arbitrary value that controls the explosion resistance of the block. For reference, stone has a resistance of 6.0, dirt has a resistance of 0.5, obsidian has a resistance of 1200, and unbreakable blocks (such as bedrock) have a hardness of 3600000.
```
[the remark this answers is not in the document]
IchHabeHunger54: As above, to my knowledge there's no known value to time conversion rate for this. Moved to method reference as well, again unsure what to do about this.

docs/blocks/index.md
IchHabeHunger54: Linking to the Minecraft wiki doesn't explain where your files go either, it just explains the general format. I put the link to the Resources section in here for now; I'm gonna try and rework Resources after this is done, I just didn't wanna do that in the same PR, as it'd get messy.

IchHabeHunger54:
I also dismissed most the feedback since I fixed most of it when rewriting ("fixed" from my point of view). If there are problems with it still, feel free to re-add feedback as needed.

docs/blocks/index.md
ChampionAsh5357: True, but it's better than nothing. Splitting it into separate PRs is probably for the best. Most of the docs need to be rewritten anyway, so I'll take what you're willing to do. I'll cover the rest once I have my hands on my new laptop.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
 Blocks
 ======
 
-Blocks are, obviously, essential to the Minecraft world. They make up all of the terrain, structures, and machines. Chances are if you are interested in making a mod, then you will want to add some blocks. This page will guide you through the creation of blocks, and some of the things you can do with them.
+Blocks are, obviously, essential to the Minecraft world. They make up all the terrain, structures, and machines. Chances are if you are interested in making a mod, then you will want to add some blocks. This page will guide you through the creation of blocks, and some of the things you can do with them.
```
ChampionAsh5357, line 4:
```suggestion
Blocks are essential to the Minecraft world. They make up all the terrain, structures, and machines. Chances are if you are interested in making a mod, then you will want to add some blocks. This page will guide you through the creation of blocks, and some of the things you can do with them.
```

Going to remove the 'obviously' because it sounds a bit patronizing.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
-----------------
+One Block to Rule Them All
+--------------------------
+
+Before we get started, it is important to understand that there is only ever one of each block in the game. A world consists of thousands of references to that one block in different locations. In other words, the same block is just displayed a lot of times.
+
+Due to this, a block should only ever be instantiated once, and that is during registration. Once the block is registered, you can then use the registered reference as needed. Consider this example (see the [Registration][registration] page if you do not know what you are looking at):
+
+```java
+public static final RegistryObject<Block> MY_BLOCK = BLOCKS.register("my_block", () -> new Block(...));
```
ChampionAsh5357, line 14: Leave a comment above saying that it's for some `DefferedRegister<Block> BLOCKS` just so that it's obvious where the variable comes from.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
+```
+
+After registering the block, all references to the new `my_block` should use this constant. For example, if you want to check if the block at a given position is a `my_block`, the code for that would look something like this:
```
ChampionAsh5357, line 17:
```suggestion
After registering the block, all references to the new `my_block` should use this constant. For example, if you want to check if the block at a given position is `my_block`, the code for that would look something like this:
```

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
+
+```java
+level.getBlockState(position).is(MyBlockRegistrationClass.MY_BLOCK.get());
```
ChampionAsh5357, line 20: Where does level come from here? This is someone's first block, so things will need to be spelled out to some degree.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
+```
+
+:::danger
+As soon as you create a `new Block()` outside registration, things can and will break!
```
ChampionAsh5357, line 24: I would make this more negative that you shouldn't create a new block outside of registration.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
+For simple blocks which need no special functionality (think cobblestone, wooden planks, etc.), the `Block` class can be used directly. To do so, during registration, instantiate `Block` with a `BlockBehaviour.Properties` parameter. This `BlockBehaviour.Properties` parameter can be created using `BlockBehaviour.Properties#of`, and it can be customized by calling its methods. The most important methods for this are:
+
+- `destroyTime` - Determines the time the block needs to be destroyed.
+- `explosionResistance` - Determines the explosion resistance of the block.
+- `sound` - Sets the sound the block makes when it is punched, broken, or placed.
+- `lightLevel` - Sets the light emission of the block. Accepts a function with a `BlockState` parameter that returns a value between 0 and 15.
+- `friction` - Sets the friction (slipperiness) of the block.
```
ChampionAsh5357, line 38: Keeping the reference values here is useful imo.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
+        .lightLevel(state -> 7)));
+```
+
+For further documentation, see the [`BlockBehaviour.Properties` reference][propertiesreference]. For more examples, or to look at the values used by Minecraft, have a look at the `Blocks` class.
 
 :::note
-Blocks have no setter for their `CreativeModeTab`. This is handled by the [`BuildCreativeModeTabContentsEvent`][creativetabs] if the block has an associated item (e.g. `BlockItem`). Furthermore, there is no setter for translation key of the block as it is generated from the registry name via `Block#getDescriptionId`.
+It is important to understand that a block in the world is not the same thing as in an inventory. What looks like a block in an inventory is actually a `BlockItem`, a special type of [item] that places a block when used. This also means that things like the creative tab or the max stack size are handled by the corresponding `BlockItem`.
+
+A `BlockItem` must be registered separately from the block. This is because a block does not necessarily need an item, for example if it is not meant to be collected (as is the case with fire, for example).
```
ChampionAsh5357, line 55: Probably should link the items page or provide an example registration here.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
 :::
 
-### Advanced Blocks
+### More Functionality
 
-Of course, the above only allows for extremely basic blocks. If you want to add functionality, like player interaction, a custom class is required. However, the `Block` class has many methods and unfortunately not every single one can be documented here. See the rest of the pages in this section for things you can do with blocks.
+Directly using `Block` only allows for very basic blocks. If you want to add functionality, like player interaction or a non-1x1x1 hitbox, a custom class that extends `Block` is required. The `Block` class has many methods that can be overridden to do different things; these are documented in the [`Block` reference][blockreference]. See also the Using blocks section below for some of the most common use cases for blocks.
```
ChampionAsh5357, line 60:
```suggestion
Directly using `Block` only allows for very basic blocks. If you want to add functionality, like player interaction or a different hitbox, a custom class that extends `Block` is required. The `Block` class has many methods that can be overridden to do different things; these are documented in the [`Block` reference][blockreference]. See also the Using blocks section below for some of the most common use cases for blocks.
```

'non-1x1x1' sounds like it can be 2x2x2 and so on, which is not the case. Better to just say different here.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
 
-Registering a Block
--------------------
+If you want to make a block that has different variants (think a slab that has a bottom, top, and double variant), you should use [blockstates]. And finally, if you want a block that stores additional data (think a chest that stores its inventory), a [block entity][blockentities] should be used. The rule of thumb here is that if you have a finite amount of states, use blockstates, and if you have an infinite or near-infinite amount of states, use a block entity.
```
ChampionAsh5357, line 62: Again, this brings into an issue of how many states. Imagine if I had a slab which could be rotated in 24 different directions and had a bottom, top, and double. That is 72 states, which is all stored in memory. There comes a point where multiple finite amount of states will overflow the memory buffer.

docs/blocks/index.md
```diff
@@ -1,51 +1,169 @@
...
----------------
+In several situations, multiple methods of `Block` are used at different times. The following subsections list the most common block-related pipelines. Unless specified otherwise, all methods are called on both logical sides.
 
-For information about block properties, such as those used for vanilla blocks like fences, walls, and many more, see the section on [blockstates].
+### Placing a Block
 
-[sounds]: ../gameeffects/sounds.md
-[creativetabs]: ../items/index.md#creative-tabs
-[registering]: ../concepts/registries.md#methods-for-registering
+Block placement logic is called from `BlockItem#useOn` (or some subclass's implementation thereof, such as in `PlaceOnWaterBlockItem`, which is used for lily pads). For more information on how the game gets there, see the [Interaction Pipeline][interactionpipeline].
```
ChampionAsh5357, line 79: Example here.

docs/blocks/reference.md
```diff
@@ -0,0 +1,220 @@
+Method Reference
```
ChampionAsh5357, line 1: So, I think this is amazing, but I also think it shouldn't be in the docs. This should be within Parchment or NeoForge itself since you are literally providing a Javadoc here. If there are methods that are replaced by NeoForge, then it would make sense to have a brief mention here with an example.

docs/blocks/states.md
```diff
@@ -1,96 +1,127 @@
...
+A good rule of thumb is: **if it has a different name, it should be a separate block**.  An example is making chair blocks: the direction of the chair should be a property, while the different types of wood should be separated into different blocks. So you'd have one chair block for each wood type, and each chair block has four blockstates (one for each direction).
 
-Not all blocks and situations require the usage of `BlockState`; only the most basic properties of a block should be put into a `BlockState`, and any other situation is better off with having a `BlockEntity` or being a separate `Block`. Always consider if you actually need to use blockstates for your purposes.
+### Blockstates vs. [Block Entities][blockentity]
 
-:::note
-A good rule of thumb is: **if it has a different name, it should be a separate block**.
-:::
+Here, the rule of thumb is: **if you have a finite amount of states, use a blockstate, if you have an infinite or near-infinite amount of states, use a block entity.** Block entities can store arbitrary amounts of data, but are slower than blockstates.
```
ChampionAsh5357, line 43: Gonna bring this up again where there is an edge case with many small finite states overflowing the buffer.

docs/blocks/states.md
```diff
@@ -1,96 +1,127 @@
...
-  this.stateDefinition.any()
-    .setValue(FACING, Direction.NORTH)
-    .setValue(OPEN, false)
-    .setValue(HINGE, DoorHingeSide.LEFT)
-    .setValue(POWERED, false)
-    .setValue(HALF, DoubleBlockHalf.LOWER)
-);
+public class LeverBlock extends Block {
+    public static final DirectionProperty FACING = BlockStateProperties.FACING;
+    public static final BooleanProperty POWERED = BlockStateProperties.POWERED;
```
ChampionAsh5357, line 80: Probably should have a comment that this just makes it more visible which properties are used in the block.

docs/blocks/states.md
```diff
@@ -1,96 +1,127 @@
...
+        super(pProperties);
+        registerDefaultState(stateDefinition.any() // stateDefinition.any() returns a random BlockState from an internal set, we don't care because we're setting all values ourselves anyway
+                .setValue(FACING, Direction.NORTH)
+                .setValue(POWERED, false)
+        );
+    }
+
+    @Override
+    protected void createBlockStateDefinition(StateDefinition.Builder<Block, BlockState> pBuilder) {
+        pBuilder.add(FACING, POWERED);
```
ChampionAsh5357, line 92: Comment here saying this is where the properties are defined on the block state.

docs/blocks/states.md
```diff
@@ -1,96 +1,127 @@
...
+        // code that determines which state will be used when placing down this block, depending on the BlockPlaceContext
+    }
+}
 ```
 
-If you wish to change what `BlockState` is used when placing your block, you can overwrite `Block#getStateForPlacement(BlockPlaceContext)`. This can be used to, for example, set the direction of your block depending on where the player is standing when they place it.
+Using Blockstates
+-----------------
+
+To go from `Block` to `BlockState`, call `Block#defaultBlockState()`.
```
ChampionAsh5357, line 106: Probably a bit more explanation here on default block state is what is registered in the constructor.

docs/blocks/states.md
```diff
@@ -1,96 +1,127 @@
...
+
+You can get the value of a property by calling `BlockState#getValue(Property<?>)`, passing it the property you want to get the value of. Reusing our lever example, this would look something like `Direction direction = leverBlockState.getValue(LeverBlock.FACING);`
```
ChampionAsh5357, line 108: Move the last line to a code block. It looks a bit weird in the doc.

docs/blocks/states.md
```diff
@@ -1,96 +1,127 @@
...
+
+If you want to get a `BlockState` with a different set of values, simply call `BlockState#setValue(Property<T>, T)` on an existing block state with the property and its value. With our lever, this goes something like `leverBlockState = leverBlockState.setValue(LeverBlock.FACING, Direction.SOUTH);`
```
ChampionAsh5357, line 110: Move the last line to a code block. It looks a bit weird in the doc.

docs/blocks/states.md
```diff
@@ -1,96 +1,127 @@
...
-Because all possible `BlockState`s are generated at startup, you are free and encouraged to use the reference equality operator (`==`) to check if two `BlockState`s are equal.
+To set a `BlockState` in the level, use `Level#setBlock(BlockPos, BlockState, int)`.
 
-Using `BlockState`'s
----------------------
+The `int` parameter deserves some extra explanation, as its meaning is not immediately obvious. It denotes what is known as update flags.
 
-You can get the value of a property by calling `BlockState#getValue(Property<?>)`, passing it the property you want to get the value of.
-If you want to get a `BlockState` with a different set of values, simply call `BlockState#setValue(Property<T>, T)` with the property and its value.
+To help setting the update flags correctly, there are a number of `int` constants in `Block`, prefixed with `UPDATE_`. Examples include `Block.UPDATE_NEIGHBORS` or `Block.UPDATE_CLIENTS`. These constants can be added together (for example `Block.UPDATE_NEIGHBORS + Block.UPDATE_CLIENTS`) if you wish to combine them. If you're unsure which one to use, use `Block.UPDATE_ALL`.
```
ChampionAsh5357, line 124: Probably best to explain what all does since it doesn't really update 'all'.

ChampionAsh5357, line 124: I think either works, but traditionally, flags are set together using `|`

docs/items/interactionpipeline.md
```diff
@@ -0,0 +1,115 @@
+The Interaction Pipeline
```
ChampionAsh5357, line 1: Again, I love the detail, but this is getting into Javadoc territory, where it probably would be better to put this into Parchment or NeoForge instead of having it here.

ChampionAsh5357 (changes requested):
Wow, this has improved so much. In general, I think the detail is great, but there are instances where you are straight up defining the method. That's fine when comparing between vanilla and NeoForge, but I believe most of those descriptions should go into Parchment or NeoForge itself. The interaction workflow meanwhile, can probably stick around.

Anything else I still need to think on what makes sense in the docs. I'm now marking this as the issue to resolve the block portion of #20

docs/blocks/index.md
IchHabeHunger54: Item page is already linked.

docs/blocks/reference.md
IchHabeHunger54: Removed from this PR, will PR to Parchment once I find the time.

docs/blocks/index.md
IchHabeHunger54: Do we have a definitive answer for this though? I'm not sure what the generally agreed-upon answer is here.

docs/blocks/states.md
IchHabeHunger54: Same as above, I don't know what the line is here.

docs/items/interactionpipeline.md
IchHabeHunger54: While I agree that the method reference should be Javadocs, I put this in here on purpose because this isn't immediately clear just by looking at the code, even for the methods where Javadocs are already present. Finding out the exact order in which things are called requires looking through some fairly dense code in `MultiPlayerGameMode` and `Minecraft`, which the average modder shouldn't be required to understand in my opinion.

docs/blocks/states.md
IchHabeHunger54: Being completely honest, I don't exactly know what each of these do myself. Help would be appreciated for this one.

docs/blocks/index.md
ChampionAsh5357: There's not a general answer for this since it varies depending on the complexity of the block. My general sense is that if you need more than 8-9 bits of data (256 to 512 states), then it should be a block entity.

docs/items/interactionpipeline.md
ChampionAsh5357: Fair enough. It just felt like when I was reading it that it was a giant wall of text, making it hard to consume. I'm not sure what's best to do about this without losing the quality.

docs/blocks/states.md
ChampionAsh5357: They are sort of self explanatory, but all of the logic is in `Level#markAndNotifyBlock`, which I believe is Forge added to split the logic.

`UPDATE_NEIGHBORS` basically update the surrounding six directions by calling `#neighborChanged`. From there it additionally calls `#onNeighborChange` and `#isRedstoneConductor`, if `#hasAnalogOutputSignal` return true. If the redstone conductor check succeeds, then `#getWeakChanges` is called to determine whether the block two away from the updating one should call `#neighborChanged`.

As for `UPDATE_CLIENTS`, it sends a packet on the next tick containing the blocks to change on the client.

As an additional check, if `UPDATE_KNOWN_SHAPE` isn't set and the depth is still greater than 0, it essentially recursively checks the surrounding blocks and performs updates until everything has updated. For reference, the flags passed in to the update shapes method zero out `UPDATE_NEIGHBORS`, `UPDATE_CLIENTS`, `UPDATE_SUPPRESS_DROPS`, and `UPDATE_MOVE_BY_PISTON`. While this means that the blocks themselves won't be sent through a packet on the client, they are rerendered immediately.

`UPDATE_ALL` combines everything above.

docs/blocks/index.md
```diff
@@ -1,51 +1,179 @@
...
 
-- `strength` - The hardness controls the time it takes to break the block. It is an arbitrary value. For reference, stone has a hardness of 1.5, and dirt 0.5. If the block should be unbreakable a hardness of -1.0 should be used, see the definition of `Blocks#BEDROCK` as an example. The resistance controls the explosion resistance of the block. For reference, stone has a resistance of 6.0, and dirt 0.5.
-- `sound` - Controls the sound the block makes when it is punched, broken, or placed. Requires a `SoundType` argument, see the [sounds] page for more details.
-- `lightLevel` - Controls the light emission of the block. Takes a function with a `BlockState` parameter that returns a value from zero to fifteen.
-- `friction` - Controls how slippery the block is. For reference, ice has a slipperiness of 0.98.
+- `destroyTime` - Determines the time the block needs to be destroyed.
+    - Stone has a destroy time of 1.5, dirt has 0.5, obsidian has 50, and bedrock has -1 (unbreakable).
+- `explosionResistance` - Determines the explosion resistance of the block.
+    - Stone has an explosion resistance of 6.0, dirt has 0.5, obsidian has 1200, and bedrock has 3600000.
```
ChampionAsh5357, line 41: Commas to separate numbers.

docs/blocks/index.md
ChampionAsh5357: Ah, my bad, I missed it.

ChampionAsh5357 (changes requested):
I don't really see any other issues besides the comments I mentioned previously that were unresolved. I will need to check over the logic flow at some point, but I think it may be useful to get opinions from some of the community to see if the list format is understandable.

Granted, I'm not going to wait forever, so I'll set a soft-deadline for November 11th, baring any major changes. If everything is till fine, then this will be merged.

docs/blocks/index.md
```diff
@@ -1,51 +1,179 @@
...
+
+```java
+//level.getBlockState(position) returns the blockstate placed in the given level (world) at the given position
+level.getBlockState(position).is(MyBlockRegistrationClass.MY_BLOCK.get());
+```
+
+This approach also has the convenient side effect that `block1 == block2` works and should be used instead of Java's `equals` method.
+
+:::danger
+Do not call `new Block()` outside registration! As soon as you do that, things can and will break!
```
Matyrobbrt, line 28: This doesn't really explain anything. Yes, you shouldn't (and usually can't) create blocks outside of registration, but *why*. I would appreciate if it states that
- blocks must be created while registries are unfrozen
- all created blocks must be registered by when the registries are frozen again (and if they aren't, the error is usually cryptic, especially for "normal" blocks) 
- if you manage to have an unregistered dangling block instance, the game will consider it air during syncing / saving

Matyrobbrt (changes requested):

docs/blocks/index.md
IchHabeHunger54: I'm a bit confused? Do you want me to use 1,5 instead of 1.5? If so, why?

ChampionAsh5357: 1200 -> 1,200; 3600000 -> 3,600,000

IchHabeHunger54: oh, for large numbers. sorry mb

docs/blocks/index.md
IchHabeHunger54: Added the rough estimate.

docs/blocks/states.md
IchHabeHunger54: Added (hopefully correctly xd)

docs/blocks/index.md
IchHabeHunger54: Added.

docs/blocks/index.md
```diff
@@ -1,51 +1,182 @@
...
+```
+
+After registering the block, all references to the new `my_block` should use this constant. For example, if you want to check if the block at a given position is `my_block`, the code for that would look something like this:
+
+```java
+//level.getBlockState(position) returns the blockstate placed in the given level (world) at the given position
+level.getBlockState(position).is(MyBlockRegistrationClass.MY_BLOCK.get());
+```
+
+This approach also has the convenient side effect that `block1 == block2` works and should be used instead of Java's `equals` method.
```
ChampionAsh5357, line 25: This is not really a side effect. Since each registered block is effectively a singleton instance, the two objects are always exactly the same. As for the `#equals` method, it can still be used, but it's pointless since we already know the objects are the same instance.

docs/blocks/index.md
```diff
@@ -1,51 +1,182 @@
...
+
+:::danger
+Do not call `new Block()` outside registration! As soon as you do that, things can and will break:
+
+- Blocks must be created while registries are unfrozen. NeoForge unfreezes registries for you and freezes them later, so registration is your time window to create blocks.
+- If you try to create and/or register a block when registries are frozen again, the game will crash. On top of that, error messages are usually cryptic here, so it's best to avoid them.
```
ChampionAsh5357, line 31: I would really say cryptic, it's just reports a `null` block, which refers to every single unregistered block instance.

docs/blocks/index.md
```diff
@@ -1,51 +1,182 @@
...
+- `lightLevel` - Sets the light emission of the block. Accepts a function with a `BlockState` parameter that returns a value between 0 and 15.
+    - For example, glowstone uses `state -> 15`, and torches use `state -> 14`.
+- `friction` - Sets the friction (slipperiness) of the block.
+    - Default value is 0.6. Ice uses 0.98.
 
-All these methods are *chainable* which means you can call them in series. See the `Blocks` class for examples of this.
+So for example, a simple implementation would look something like this:
+
+```java
+public static final RegistryObject<Block> MY_BETTER_BLOCK = BLOCKS.register("my_better_block", () -> new Block(BlockBehaviour.Properties.of()
```
ChampionAsh5357, line 56: Probably should split this up so it's not a scrollable element by default on a standard screen.

docs/blocks/index.md
```diff
@@ -1,51 +1,182 @@
...
+
+A `BlockItem` must be registered separately from the block. This is because a block does not necessarily need an item, for example if it is not meant to be collected (as is the case with fire, for example).
 :::
 
-### Advanced Blocks
+### More Functionality
 
-Of course, the above only allows for extremely basic blocks. If you want to add functionality, like player interaction, a custom class is required. However, the `Block` class has many methods and unfortunately not every single one can be documented here. See the rest of the pages in this section for things you can do with blocks.
+Directly using `Block` only allows for very basic blocks. If you want to add functionality, like player interaction or a different hitbox, a custom class that extends `Block` is required. The `Block` class has many methods that can be overridden to do different things; see the classes `Block`, `BlockBehaviour` and `IForgeBlock` for more information. See also the Using blocks section below for some of the most common use cases for blocks.
```
ChampionAsh5357, line 73: Link to using blocks?

docs/blocks/index.md
```diff
@@ -1,51 +1,182 @@
...
+finishingStage();
+```
+
+The following subsections further break down these stages into actual method calls.
+
+#### The "Initiating" Stage
+
+- `InputEvent.InteractionKeyMappingTriggered` is fired with the left mouse button and the main hand. If the event is canceled, the pipeline ends.
+- Several prerequisites are checked, for example that you are not in spectator mode, that all required feature flags for the `ItemStack` in your main hand are enabled or that the block in question is not outside the world border. If at least one of these checks fails, the pipeline ends.
+- `PlayerInteractEvent.LeftClickBlock` is fired. If the event is canceled, the pipeline ends.
```
ChampionAsh5357, line 132: Correct me if I'm wrong, but this has some differing behavior depending when you intercept this even on the client or server. If on the client, it will stop the packet from being sent to the server altogether which 'continues' the stage to some degree.

This comment applies to all the `PlayerInteractEvent`s.

docs/blocks/index.md
```diff
@@ -1,51 +1,182 @@
...
+    - `Block#getDestroyProgress` returns a float value between 0 and 1, representing how much the destroy progress counter should be increased every tick.
+- The progress overlay (cracking texture) is updated accordingly.
+- If the destroy progress is greater than 1.0 (i.e. completed, i.e. the block should be broken), the "mining" stage is exited and the "actually breaking" stage is entered.
+
+#### The "Actually Breaking" Stage
+
+- `Item#onBlockStartBreak` is called. If it returns `true`, the pipeline moves to the "finishing" stage.
+- Server-only: `IForgeBlock#canHarvestBlock` is called.
+- `Block#onDestroyedByPlayer` is called. If it returns `false`, the pipeline moves to the "finishing" stage. In that `Block#onDestroyedByPlayer` call:
+    - `Block#playerWillDestroy` is called.
```
ChampionAsh5357, line 148: I'm not sure whether the player has control of this method, but for the ones they do, you might want to make a comment whether any intercepting logic should be checked on the server, client, or both since information can differ.

docs/blocks/index.md
```diff
@@ -1,51 +1,182 @@
...
+
+#### The "Finishing" Stage
+
+- The internal destroy progress counter is reset.
+
+### Ticking and Random Ticking
+
+Like many other things, blocks are ticked (updated) every tick, which is 1 / 20 of a second, or 50 milliseconds.
+
+There are two methods for ticking: `Block#tick` and `Block#animateTick`. The difference is that `Block#tick` is called only on the server, whereas `Block#animateTick` is called only on the client. For example, redstone updates are processed in `Block#tick`, while the fire particles of torches are spawned in `Block#animateTick`. Override either of them to add your server/client ticking behavior.
```
ChampionAsh5357, line 164: Is `#tick` really called every tick? I don't think that's the case. I thought this is only called when someone schedules a tick update through the scheduler system.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
 
-In Minecraft 1.8 and above, the metadata system, along with the block ID system, was deprecated and eventually replaced with the **block state system**. The block state system abstracts out the details of the block's properties from the other behaviors of the block.
+If your block does not have any blockstate properties defined, it still has exactly one blockstate - that is the one without any properties, since there are no properties to specify. This can be denoted as `minecraft:oak_planks[]` or simply `minecraft:oak_planks`.
 
-Each *property* of a block is described by an instance of `Property<?>`. Examples of block properties include instruments (`EnumProperty<NoteBlockInstrument>`), facing (`DirectionProperty`), poweredness (`Property<Boolean>`), etc. Each property has the value of the type `T` parametrized by `Property<T>`.
+As with blocks, every `BlockState` exists exactly once in memory. This has the convenient side effect that `==` can and should be used to compare `BlockState`s. `BlockState` is also a final class, meaning it cannot be extended. **Any functionality goes in the corresponding [Block][block] class!**
```
ChampionAsh5357, line 32: Again, it's not really a side effect, immutable singletons make identities.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
 
-:::note
-A good rule of thumb is: **if it has a different name, it should be a separate block**.
-:::
+Blockstates and block entities can be used in conjunction with one another. For example, the chest uses blockstate properties for things like the direction, whether it is opened or not, or becoming a double chest is handled by blockstate properties, while storing the inventory or interacting with hoppers is handled by a block entity.
 
-An example is making chair blocks: the *direction* of the chair should be a *property*, while the different *types of wood* should be separated into different blocks.
-An "Oak Chair" facing east (`oak_chair[facing=east]`) is different from a "Spruce Chair" facing west (`spruce_chair[facing=west]`).
+There is no definitive answer to the question "How many states are too much for a blockstate?", but the generally agreed upon answer is that if you need more than 8-9 bits of data (i.e. more than a few hundred states), use a block entity instead.
```
ChampionAsh5357, line 47: Probably should say 'the authors recommend' since no one has generally agreed on this.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
+
+Every block will also have a default state. If nothing else is specified, the default state uses the default value of every property. You can change the default state by calling the `Block#registerDefaultState(BlockState)` method from your constructor.
+
+If you wish to change which `BlockState` is used when placing your block, override `Block#getStateForPlacement(BlockPlaceContext)`. This can be used to, for example, set the direction of your block depending on where the player is standing or looking when they place it.
+
+To further illustrate this, this is what the relevant bits of the `LeverBlock` class look like:
+
+```java
+public class LeverBlock extends Block {
+    // Note: It is possible to directly use the values in BlockStateProperties instead of referencing them here again. However, for the sake of simplicity and readability, it is recommended to add constants like this.
```
ChampionAsh5357, line 81: Multiline this to avoid unnecessary wrapping when page full screen.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
+    public static final DirectionProperty FACING = BlockStateProperties.FACING;
+    public static final BooleanProperty POWERED = BlockStateProperties.POWERED;
+
+    public LeverBlock(BlockBehaviour.Properties pProperties) {
+        super(pProperties);
+        registerDefaultState(stateDefinition.any() // stateDefinition.any() returns a random BlockState from an internal set, we don't care because we're setting all values ourselves anyway
```
ChampionAsh5357, line 87: Multiline this to avoid unnecessary wrapping when page full screen.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
+                .setValue(FACING, Direction.NORTH)
+                .setValue(POWERED, false)
+        );
+    }
+
+    @Override
+    protected void createBlockStateDefinition(StateDefinition.Builder<Block, BlockState> pBuilder) {
+        pBuilder.add(FACING, POWERED); // this is where the properties are actually added to the state
```
ChampionAsh5357, line 95: Multiline this to avoid unnecessary wrapping when page full screen.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
+    }
+
+    @Override
+    @Nullable
+    public BlockState getStateForPlacement(BlockPlaceContext pContext) {
+        // code that determines which state will be used when placing down this block, depending on the BlockPlaceContext
```
ChampionAsh5357, line 101: Multiline this to avoid unnecessary wrapping when page full screen.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
+
+Using Blockstates
+-----------------
+
+To go from `Block` to `BlockState`, call `Block#defaultBlockState()`. The default blockstate can be changed through `Block#registerDefaultState`, as described above.
+
+You can get the value of a property by calling `BlockState#getValue(Property<?>)`, passing it the property you want to get the value of. Reusing our lever example, this would look something like this:
+
+```java
+Direction direction = leverBlockState.getValue(LeverBlock.FACING);
```
ChampionAsh5357, line 114: Comment to reiterate that `FACING` references the `Direction` property on the `BlockState`

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
 
-Using `BlockState`'s
----------------------
+To help setting the update flags correctly, there are a number of `int` constants in `Block`, prefixed with `UPDATE_`. These constants can be bitwise-ORed together (for example `Block.UPDATE_NEIGHBORS | Block.UPDATE_CLIENTS`) if you wish to combine them.
 
-You can get the value of a property by calling `BlockState#getValue(Property<?>)`, passing it the property you want to get the value of.
-If you want to get a `BlockState` with a different set of values, simply call `BlockState#setValue(Property<T>, T)` with the property and its value.
+- `Block.UPDATE_NEIGHBORS` sends an update to the neighboring blocks. More specifically, it calls `Block#neighborChanged`, which calls a number of methods, most of which are redstone-related in some way.
+- `Block.UPDATE_CLIENTS` syncs the block update to the client.
+- `Block.UPDATE_INVISIBLE` explicitly does not update. This also overrules `Block.UPDATE_CLIENTS`, causing the update to not be synced.
```
ChampionAsh5357, line 139: It does not update on the client. The server is always updated.

docs/blocks/states.md
```diff
@@ -1,96 +1,149 @@
...
+- `Block.UPDATE_IMMEDIATE` forces the update immediately, instead of on the next tick.
```
ChampionAsh5357, line 140: Sort of? This one is a bit hard to explain. Just because this is updated immediately doesn't mean that the changes will be visibly applied on the same tick. You might want to take a gander through the method in question again.

ChampionAsh5357 (changes requested):
More things I noticed while reading the docs on the rendered page.

docs/blocks/index.md
IchHabeHunger54: I went by Maty's statement here, I never encountered such an error myself. Will correct.

docs/blocks/index.md
IchHabeHunger54: Yes, it is called every tick through ServerLevel#tick.

docs/blocks/states.md
IchHabeHunger54: This is well-hidden in LevelRenderer#blockChanged, which is why I couldn't find it.

docs/blocks/index.md
ChampionAsh5357: So, I took a look, the method is *passed through* a tick block method which only ticks a block that has been scheduled on that specific tick. Think about it, if blocks tick every tick, what is the point of a block entity ticker method?

embeddedt:
I am not sure what's happening (could be a bug on my end), but it seems like many of the issues raised in the review comments are still present in the diff view on GitHub, despite the comment being marked as resolved.

IchHabeHunger54:
Yes, I forgot to push.

docs/blocks/index.md
IchHabeHunger54: I misinterpreted that specific code then.

docs/blocks/index.md
IchHabeHunger54: The modder should always do behavior the same way on both sides (unless the method is only called on one side to begin with, of course).

docs/blocks/index.md
```diff
@@ -1,51 +1,186 @@
...
+```java
+//BLOCKS is a DeferredRegister<Block>
+public static final RegistryObject<Block> MY_BLOCK = BLOCKS.register("my_block", () -> new Block(...));
+```
+
+After registering the block, all references to the new `my_block` should use this constant. For example, if you want to check if the block at a given position is `my_block`, the code for that would look something like this:
+
+```java
+//level.getBlockState(position) returns the blockstate placed in the given level (world) at the given position
+level.getBlockState(position).is(MyBlockRegistrationClass.MY_BLOCK.get());
```
Matyrobbrt, line 22:
```suggestion
level.getBlockState(position) // returns the blockstate placed in the given level (world) at the given position
// highlight-next-line
    .is(MyBlockRegistrationClass.MY_BLOCK.get())
```

docs/blocks/index.md
```diff
@@ -1,51 +1,186 @@
...
+
+```java
+public static final RegistryObject<Block> MY_BETTER_BLOCK = BLOCKS.register(
+        "my_better_block", 
+        () -> new Block(BlockBehaviour.Properties.of()
+                .destroyTime(2.0f)
+                .explosionResistance(10.0f)
+                .sound(SoundType.GRAVEL)
+                .lightLevel(state -> 7)
+        ));
```
Matyrobbrt, line 63:
```suggestion
        () -> new Block(BlockBehaviour.Properties.of()
                // highlight-start
                .destroyTime(2.0f)
                .explosionResistance(10.0f)
                .sound(SoundType.GRAVEL)
                .lightLevel(state -> 7)
                // highlight-end
        ));
```

Matyrobbrt (changes requested):

ChampionAsh5357 (changes requested):
In general, I think this is pretty much done. There could be more examples, but from what I read, barring my general disposition, I believe this is good.

docs/items/interactionpipeline.md
```diff
@@ -0,0 +1,115 @@
...
+        - If the event does not deny item usage, `Item#useOn` is called. If it returns a definitive result, the pipeline ends.
+- `Item#use` is called. If it returns a definitive result, the pipeline ends.
+- The above process runs a second time, this time with the off hand instead of the main hand.
+
+Result Types
+------------
+
+There are two different types of results: `InteractionResult`s and `InteractionResultHolder<T>`s. Depending on the situation, one of the two is used.
+
+`InteractionResult` is an enum consisting of five values: `SUCCESS`, `CONSUME`, `CONSUME_PARTIAL`, `PASS` and `FAIL`. Additionally, the method `InteractionResult#sidedSuccess` is available, which returns `SUCCESS` on the server and `CONSUME` on the client.
```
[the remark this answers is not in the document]
IchHabeHunger54: I rewrote the section to mention "default" meanings and then explain the non-default behavior.

docs/blocks/states.md
```diff
@@ -1,96 +1,155 @@
...
+When to Use Blockstates
+-----------------------
 
-A unique pair can be constructed from the `Block` and a map of the `Property<?>` to their associated values. This unique pair is called a `BlockState`.
+### Blockstates vs. Separate Blocks
 
-The previous system of meaningless metadata values were replaced by a system of block properties, which are easier to interpret and deal with. Previously, a stone button which is facing east and is powered or held down is represented by "`minecraft:stone_button` with metadata `9`. Now, this is represented by "`minecraft:stone_button[facing=east,powered=true]`".
+A good rule of thumb is: **if it has a different name, it should be a separate block**.  An example is making chair blocks: the direction of the chair should be a property, while the different types of wood should be separated into different blocks. So you'd have one chair block for each wood type, and each chair block has four blockstates (one for each direction).
```
ChampionAsh5357, line 39:
```suggestion
A good rule of thumb is: **if it has a different name, it should be a separate block**. An example is making chair blocks: the direction of the chair should be a property, while the different types of wood should be separated into different blocks. So you'd have one chair block for each wood type, and each chair block has four blockstates (one for each direction).
```

ChampionAsh5357 (dismissed):
Small redundant space, but that's all I've noticed. Everything else seems correct from my understanding. Granted, I cannot look at the codebase rn, so I'll leave that to @Matyrobbrt and @XFactHD to double check the logic.

ChampionAsh5357 (dismissed):
Yep, gonna give maty and xfact until the 20th if they want to change anything. Otherwise, I'll merge it.

docs/items/interactionpipeline.md
```diff
@@ -0,0 +1,71 @@
...
+        - `Entity#interact` is called **on the entity you are looking at**. If it returns a definitive result, the pipeline ends.
+            - If you want to add behavior for your own entity, override this method. If you want to add behavior for a vanilla entity, use the event.
+            - For `Mob`s, the override of `Entity#interact` handles things like leashing and spawning babies when the `ItemStack` in your main hand is a spawn egg, and then defers mob-specific handling to `Mob#mobInteract`. The rules for results for `Entity#interact` apply here as well.
+        - If the entity you are looking at is a `LivingEntity`, `Item#interactLivingEntity` is called on the `ItemStack` in your main hand. If it returns a definitive result, the pipeline ends.
+    - If you are looking at a block that is within your reach and not outside the world border:
+        - `PlayerInteractEvent.RightClickBlock` is fired. If the event is canceled, the pipeline ends. You may also specifically deny only block or item usage in this event.
+        - `IForgeItem#onItemUseFirst` is called. If it returns a definitive result, the pipeline ends.
```
XFactHD, line 26:
```suggestion
        - `IItemExtension#onItemUseFirst` is called. If it returns a definitive result, the pipeline ends.
```

docs/items/interactionpipeline.md
```diff
@@ -0,0 +1,71 @@
...
+- `InteractionResult.SUCCESS` (or `InteractionResultHolder#success` where needed) should be used if the operation should be considered successful, and you want the arm to swing, but only on one side. Only use this if you want to return a different value on the other logical side for whatever reason. The pipeline will end.
+- `InteractionResult.CONSUME` (or `InteractionResultHolder#consume` where needed) should be used if the operation should be considered successful, but you do not want the arm to swing. The pipeline will end.
+- `InteractionResult.CONSUME_PARTIAL` is mostly identical to `InteractionResult.CONSUME`, the only difference is in its usage in [`Item#useOn`][itemuseon].
+- `InteractionResult.FAIL` (or `InteractionResultHolder#fail` where needed) should be used if the item functionality should be considered failed and no further interaction should be performed. The pipeline will end. This should only be used in `Item#useOn` and `Item#use`, all other implementations should use `InteractionResult.PASS` to allow item behavior to still be called.
```
XFactHD, line 47: > all other implementations should use `InteractionResult.PASS` to allow item behavior to still be called

This is not necessarily true, there are valid reasons to cancel the whole pipeline by returning "fail" from `Block#use()` for example

docs/items/interactionpipeline.md
```diff
@@ -0,0 +1,71 @@
...
+- `InteractionResult.PASS` (or `InteractionResultHolder#pass` where needed) should be used if the operation should be considered neither successful nor failed. The pipeline will continue. This is the default behavior (unless otherwise specified).
+
+Some methods have special behavior or requirements, which are explained in the below chapters.
+
+`IForgeItem#onItemUseFirst`
```
XFactHD, line 52:
```suggestion
`IItemExtension#onItemUseFirst`
```

XFactHD (changes requested):
Overall looks good to me, just needs a few adjustments

IchHabeHunger54:
I decided to use `EndPortalFrameBlock` as an example rather than `AbstractFurnaceBlock`, as it doesn't have any block entity things going on that might lead to confusion when new modders decide to look there for an example.

docs/blocks/states.md
```diff
@@ -1,96 +1,153 @@
...
-    .setValue(POWERED, false)
-    .setValue(HALF, DoubleBlockHalf.LOWER)
-);
+public class EndPortalFrameBlock extends Block {
+    // Note: It is possible to directly use the values in BlockStateProperties instead of referencing them here again.
+    // However, for the sake of simplicity and readability, it is recommended to add constants like this.
+    public static final DirectionProperty FACING = BlockStateProperties.FACING;
+    public static final BooleanProperty EYE = BlockStateProperties.EYE;
+
+    public LeverBlock(BlockBehaviour.Properties pProperties) {
```
XFactHD, line 82:
```suggestion
    public EndPortalFrameBlock(BlockBehaviour.Properties pProperties) {
```

XFactHD (changes requested):

XFactHD (dismissed):

docs/blocks/index.md
```diff
@@ -1,51 +1,206 @@
...
+--------------------------
 
-### Basic Blocks
+Before we get started, it is important to understand that there is only ever one of each block in the game. A world consists of thousands of references to that one block in different locations. In other words, the same block is just displayed a lot of times.
+
+Due to this, a block should only ever be instantiated once, and that is during registration. Once the block is registered, you can then use the registered reference as needed. Consider this example (see the [Registration][registration] page if you do not know what you are looking at):
+
+```java
+//BLOCKS is a DeferredRegister<Block>
+public static final RegistryObject<Block> MY_BLOCK = BLOCKS.register("my_block", () -> new Block(...));
```
Matyrobbrt, line 15: Any reason this doesn't use the new DR for blocks? Also `RegistryObject` is gone now.

Matyrobbrt (changes requested):

docs/blocks/index.md
IchHabeHunger54: Written before the registry PR was merged, will adjust.

docs/blocks/index.md
```diff
@@ -54,7 +54,7 @@ For simple blocks which need no special functionality (think cobblestone, wooden
 So for example, a simple implementation would look something like this:
 
 ```java
-public static final RegistryObject<Block> MY_BETTER_BLOCK = BLOCKS.register(
+  public static final Supplier<Block> MY_BETTER_BLOCK = BLOCKS.register(
```
XFactHD, line 57:
```suggestion
  public static final DeferredBlock<Block> MY_BETTER_BLOCK = BLOCKS.register(
```
Should be a `Holder<Block>`, `DeferredHolder<Block, Block>` or a `DeferredBlock<Block>`, whichever you prefer.

Matyrobbrt (dismissed):

docs/blocks/index.md
```diff
@@ -1,51 +1,200 @@
...
+--------------------------
 
-### Basic Blocks
+Before we get started, it is important to understand that there is only ever one of each block in the game. A world consists of thousands of references to that one block in different locations. In other words, the same block is just displayed a lot of times.
+
+Due to this, a block should only ever be instantiated once, and that is during registration. Once the block is registered, you can then use the registered reference as needed. Consider this example (see the [Registration][registration] page if you do not know what you are looking at):
+
+```java
+//BLOCKS is a DeferredRegister.Blocks
+public static final Supplier<Block> MY_BLOCK = BLOCKS.register("my_block", () -> new Block(...));
```
XFactHD, line 15:
```suggestion
public static final DeferredBlock<Block> MY_BLOCK = BLOCKS.register("my_block", () -> new Block(...));
```

XFactHD (changes requested):
Final thing I missed yesterday, finally good to go from me when that's fixed

XFactHD (approved):

Matyrobbrt (approved):

ChampionAsh5357 (approved):
Looks good. Will try to merge on phone.

PR #7, opened 2023-08-04
Delete forgedev category

tmvkrpxl0:
This was requested from Maty, member of documentation team.

------------------
Preview URL: https://pr-7.neoforged-docs-previews.pages.dev

PR #22, opened 2023-10-17
Add contribution to NeoForge guidelines, ported from Forge

pandaninjas:


ChampionAsh5357:
This was removed purposely as we are currently in the process of rewriting the toolchain, so this information is out of date. It will be added back in once everything is stable.

For reference, this was done via #7.

PR #23, opened 2023-10-25
Updated contributer in readme

Ram-tyagi:
Update the contributor in the readme file


ChampionAsh5357:
The list of contributors are proudly displayed on the right sidebar of the repository, like all other Github repos. There is no need to have this directly in the README.

PR #25, opened 2023-11-02
Upgrade to Docusaurus v3

Mysterious-Dev:
All is in the title ^^.

Should be merged before #9 to test if changes for this PR is OK with Docusaurus V3.

------------------
Preview URL: https://pr-25.neoforged-docs-previews.pages.dev


ChampionAsh5357 (approved):
Everything seems to be functioning correctly, so I have no issues. I don't think the rename in the code was necessary, but it does make the js clearer.

PR #34, opened 2023-12-07
Update docs for capability rework

Technici4n:
Mostly adapted from the blog post, but keeping some of the old content where that made sense.

Beyond the two main changed files, I also fixed references to capabilities elsewhere. Some of the other pages will still need an update, but at least everything related to caps should be clean now.

------------------
Preview URL: https://pr-34.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/datastorage/attachments.md
+++ b/docs/datastorage/attachments.md
@@ -89,6 +89,7 @@ Currently, only serializable item stack attachments are synced between the clien
 This is done automatically.
 
 To sync block entity, chunk, or entity attachments to a client, you need to [send a packet to the client][network] yourself.
+For chunks, you can use `ChunkWatchEvent.Sent` to know when to send chunk data to a player.
 
 ## Copying data on player death
 By default, entity data attachments are not copied on player death.
--- a/docs/datastorage/attachments.md
+++ b/docs/datastorage/attachments.md
@@ -1,0 +1,93 @@
+# Data Attachments
+
+The data attachment system allows mods to attach and store additional data on block entities, chunks, entities, and item stacks.
+
+_To store additional level data, you can use [SavedData](saveddata)._
+
+## Creating an attachment type
+
+To use the system, you need to register an `AttachmentType`.
+The attachment type contains:
+- a default value supplier to create the instance when the data is first accessed, or to compare stacks that have the data and stacks that don't have it;
+- an optional serializer if the attachment should be persisted;
+- additional configuration options for the attachment, for example the `copyOnDeath` flag.
+
+:::danger
+If you don't provide a serializer, the attachment will not be saved to disk.
+If your data attachment disappears when you reload the world, you probably forgot to provide a serializer.
+:::
+
+There are a few ways to provide an attachment serializer: directly implementing `IAttachmentSerializer`, implementing `INBTSerializable` and using the static `AttachmentSerializer.serializable()` method to create the builder, or providing a codec to the builder. (This latter option is not recommended for item stacks due to relatively slowness).
+
+In any case, we recommend using a `DeferredRegister` for registration:
+```java
+// Create the DeferredRegister for attachment types
+private static final DeferredRegister<AttachmentType<?>> ATTACHMENT_TYPES = DeferredRegister.create(NeoForgeRegistries.Keys.ATTACHMENT_TYPES, MOD_ID);
+
+// Serialization via INBTSerializable
+private static final Supplier<AttachmentType<ItemStackHandler>> HANDLER = ATTACHMENT_TYPES.register(
+        "handler", () -> AttachmentType.serializable(() -> new ItemStackHandler(1)).build());
+// Serialization via codec
+private static final Supplier<AttachmentType<Integer>> MANA = ATTACHMENT_TYPES.register(
+        "mana", () -> AttachmentType.builder(() -> 0).serialize(Codec.INT).build());
+// No serialization
+private static final Supplier<AttachmentType<SomeCache>> SOME_CACHE = ATTACHMENT_TYPES.register(
+        "some_cache", () -> AttachmentType.builder(() -> new SomeCache()).build()
+);
+
+// Don't forget to register the DeferredRegister to your mod bus:
+ATTACHMENT_TYPES.register(modBus);
+```
+
+## Using the attachment type
+
+Once the attachment type is registered, it can be used on any holder object.
+Calling `getData` if no data is present will attach a new default instance.
+
+```java
+// Get the ItemStackHandler if it already exists, else attach a new one:
+ItemStackHandler stackHandler = stack.getData(HANDLER);
+// Get the current player mana if it is available, else attach 0:
+int playerMana = player.getData(MANA);
+// And so on...
+```
+
+If attaching a default instance is not desired, a `hasData` check can be added:
+```java
+// Check if the stack has the HANDLER attachment before doing anything.
+if (stack.hasData(HANDLER)) {
+    ItemStackHandler stackHandler = stack.getData(HANDLER);
+    // Do something with stack.getData(HANDLER).
+}
+```
+
+The data can also be updated with `setData`:
+```java
+// Increment mana by 10.
+player.setData(MANA, player.getData(MANA) + 10);
+```
+
+Usually, block entities and chunks need to be marked as dirty when they are modified (with `setChanged` and `setUnsaved(true)`). This is done automatically for calls to `setData`:
+```java
+chunk.setData(MANA, chunk.getData(MANA) + 10); // will call setUnsaved automatically
+```
+but if you modify some data that you obtained from `getData` then you must mark block entities and chunks as dirty explicitly:
+```java
+var mana = chunk.getData(MUTABLE_MANA);
+mana.set(10);
+chunk.setUnsaved(true); // must be done manually because we did not use setData
+```
+
+## Sharing data with the client
+Currently, only serializable item stack attachments are synced with the client.
+This is done automatically.
+
+To sync block entity, chunk, or entity attachments, you need to [send a packet to the client][network] yourself.
+
+## Copying data on player death
+By default, entity data attachments are not copied on player death.
+To automatically copy an attachment on player death, set `.copyOnDeath()` in the attachment builder.
+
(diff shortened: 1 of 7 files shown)


docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+
+The data attachment system allows mods to attach and store additional data on block entities, chunks, entities, and item stacks.
+
+_To store additional level data, you can use [SavedData](saveddata)._
+
+## Creating an attachment type
+
+To use the system, you need to register an `AttachmentType`.
+The attachment type contains:
+- a default value supplier to create the instance when the data is first accessed, or to compare stacks that have the data and stacks that don't have it;
```
ChampionAsh5357, line 11: Is this only true for stacks, or does it apply to any attached object?

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+- an optional serializer if the attachment should be persisted;
+- additional configuration options for the attachment, for example the `copyOnDeath` flag.
```
ChampionAsh5357, line 13: This probably should be explained more since these additional configurations require the serializer to be present.

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+
+:::warning
+If you don't provide a serializer, the attachment will not be saved to disk.
+(This can be intended).
```
ChampionAsh5357, line 17: This is a weird way of just saying 'If you don't want your attachment to persist its data, do not provide a serializer.'

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+If your data attachment disappears when you reload the world but you want it to persist, you probably forgot to provide a serializer.
+:::
+
+There are a few ways to provide an attachment serializer: directly implementing `IAttachmentSerializer`, implementing `INBTSerializable` and using the static `AttachmentSerializer.serializable()` method to create the builder, or providing a codec to the builder. (This latter option is not recommended for item stacks due to relatively slowness).
```
ChampionAsh5357, line 21: Latter option should be explicitly stated in a warning.

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+
+In any case, we recommend using a `DeferredRegister` for registration:
```
ChampionAsh5357, line 23: This line isn't really necessary. It's more important to state that attachment types must be registered.

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+// Serialization via codec
+private static final Supplier<AttachmentType<Integer>> MANA = ATTACHMENT_TYPES.register(
+        "mana", () -> AttachmentType.builder(() -> 0).serialize(Codec.INT).build());
+// No serialization
+private static final Supplier<AttachmentType<SomeCache>> SOME_CACHE = ATTACHMENT_TYPES.register(
+        "some_cache", () -> AttachmentType.builder(() -> new SomeCache()).build()
+);
+
+// Don't forget to register the DeferredRegister to your mod bus:
+ATTACHMENT_TYPES.register(modBus);
```
ChampionAsh5357, line 40: Comment here saying to add it within the mod's constructor.

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+Usually, block entities and chunks need to be marked as dirty when they are modified (with `setChanged` and `setUnsaved(true)`). This is done automatically for calls to `setData`:
+```java
+chunk.setData(MANA, chunk.getData(MANA) + 10); // will call setUnsaved automatically
+```
+but if you modify some data that you obtained from `getData` then you must mark block entities and chunks as dirty explicitly:
+```java
+var mana = chunk.getData(MUTABLE_MANA);
+mana.set(10);
+chunk.setUnsaved(true); // must be done manually because we did not use setData
+```
```
ChampionAsh5357, line 80: This should probably be in an important block since it could otherwise be glossed over.

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Sharing data with the client
+Currently, only serializable item stack attachments are synced with the client.
```
ChampionAsh5357, line 83: Does this also fix the bug with item stack in the creative menu involving the weirdness when an attachment is present on the client and needs to be synced back to the server?

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+This is done automatically.
+
+To sync block entity, chunk, or entity attachments, you need to [send a packet to the client][network] yourself.
```
ChampionAsh5357, line 86: Just for confirmation, these can be sent any time the attachment is accessed?

docs/datastorage/attachments.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Copying data on player death
+By default, entity data attachments are not copied on player death.
+To automatically copy an attachment on player death, set `.copyOnDeath()` in the attachment builder.
+
+More complex handling can be implemented via `PlayerEvent.Clone` by reading the data from the original entity and assigning it to the new entity. In this event, the `#isWasDeath` method can be used to distinguish between respawning after death and returning from the End. This is important because the data will already exist when returning from the End, so care has to be taken to not duplicate values in this case.
```
ChampionAsh5357, line 92: Example should be provided here.

docs/datastorage/capabilities.md
```diff
@@ -1,177 +1,318 @@
...
+```java
+modBus.addListener(RegisterCapabilitiesEvent.class, event -> {
+    event.registerItem(
+        Capabilities.FluidHandler.ITEM,
+        (stack, ctx) -> new MyCustomFluidBucketWrapper(stack),
+        // blocks to register for
+        MY_CUSTOM_BUCKET);
+}, EventPriority.HIGH); // use HIGH priority to register before NeoForge!
+```
+See `CapabilityHooks` for a list of the providers registered by NeoForge itself.
```
ChampionAsh5357, line 316: Link!

docs/datastorage/capabilities.md
```diff
@@ -1,177 +1,318 @@
...
+    if (item.getClass() == BucketItem.class) {
+        event.registerItem(Capabilities.FluidHandler.ITEM, (stack, ctx) -> new FluidBucketWrapper(stack), item);
+    }
+}
+```
 
-This can be done via `PlayerEvent$Clone` by reading the data from the original entity and assigning it to the new entity. In this event, the `#isWasDeath` method can be used to distinguish between respawning after death and returning from the End. This is important because the data will already exist when returning from the End, so care has to be taken to not duplicate values in this case.
+Providers are asked for a capability in the order that they are registered.
+Should you want to run before a provider that NeoForge already registers for one of your objects,
+register your `RegisterCapabilitiesEvent` handler with a higher priority.
```
ChampionAsh5357, line 305: When would this matter?

ChampionAsh5357 (changes requested):
Bunch of general questions because my lack of understanding of the underlying system. I have avoided commenting on syntax changes as much as possible as per your request. They will be addressed at a later point in time since they don't affect the user experience.

docs/datastorage/capabilities.md
```diff
@@ -1,177 +1,318 @@
...
+        () -> onCapInvalidate() // invalidation listener
+);
 ```
 
-### @AutoRegisterCapability
-
-A capability is registered using `@AutoRegisterCapability` by annotating the capability type.
+## Block capability invalidation
+:::info
+Invalidation is exclusive to block capabilities. Entity and item stack capabilities do not need to be invalidated.
```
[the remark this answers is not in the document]
Technici4n: For entities and items, the cap is only valid right after the query and should not be cached

docs/datastorage/capabilities.md
```diff
@@ -1,177 +1,318 @@
...
+To give a more concrete example, here is how one might query an `IItemHandler` capability for a block, from the `Direction.NORTH` side:
+```java
+IItemHandler handler = level.getCapability(Capabilities.ItemHandler.BLOCK, pos, Direction.NORTH);
+if (handler != null) {
+    // Use the handler for some item-related operation.
+}
+```
 
-As mentioned, attaching capabilities to existing providers, `Level`s, and `LevelChunk`s can be done using `AttachCapabilitiesEvent`. The same event is used for all objects that can provide capabilities. `AttachCapabilitiesEvent` has 5 valid generic types providing the following events:
+## Block capability caching
```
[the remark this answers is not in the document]
Technici4n: This is not written on the page but the cache will not query unloaded chunks. Of course, once the chunk loads the cache will receive a notification and try to query again. Not sure if that answers the question?

docs/datastorage/capabilities.md
```diff
@@ -1,177 +1,318 @@
...
+// whenever a capability changes, appears, or disappears:
+level.invalidateCapabilities(pos);
 ```
 
-Persisting LevelChunk and BlockEntity capabilities
---------------------------------------------
+NeoForge already handles common cases such as chunk load/unloads and block entity creation/removal,
+but other cases need to be handled explicitly by modders.
+For example, modders must invalidate capabilities in the following cases:
+- If the configuration of a capability-providing block entity changes.
```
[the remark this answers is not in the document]
Technici4n: In the sense of a machine configuration changing in such a way that previously returned capabilities are no longer valid, or that new capabilities have become available. Configuration isn't the clearest yeah

docs/datastorage/attachments.md
ChampionAsh5357: I think it's a good idea to have if someone has such a use case. The documentation is meant to be for people who are beginners or experienced modders, so if we mention something, we should have any corollaries necessary.

docs/datastorage/capabilities.md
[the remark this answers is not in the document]
Technici4n: On the next query.

ChampionAsh5357 (approved):
Good enough, any formatting errors can be addressed in a future commit. @XFactHD and @Matyrobbrt, will give you until whenever Thursday to do any additional reviews or comments.

XFactHD:
Looks good to me overall, just a few minor points
- It might be worth mentioning that the fluid capability for items is slightly special due to the way buckets hold fluids
- The remark regarding why block caps are looked up from the level now should explicitly mention why (i.e. that blocks without BEs can have caps now as well)
- If the holder of a data attachment requires it, is it automatically marked dirty when a non-existent attachment is retrieved and therefore default-initialized?

Other than that: :shipit:

XFactHD (approved):

PR #37, opened 2023-12-16
Describe type + reason introduced in NeoForge 20.4.40-beta

shartte:
See https://github.com/neoforged/FancyModLoader/pull/51

------------------
Preview URL: https://pr-37.neoforged-docs-previews.pages.dev


IchHabeHunger54:
Looks good in general. Three minor nitpicks:
- Please remove the "Replaced the `mandatory` field in NeoForge 20.4.40-beta". We assume people reading the docs to be on latest NeoForge.
- AFAICT the type value must be encased in "" (this is not the case in the example).
- Please add a simple example for the reason, something like `reason="This mod is not compatible with <other mod> because both touch rendering internals in a non-compatible way."`

IchHabeHunger54 (approved):

PR #44, opened 2024-01-03
Remove version 1.19.x

Imzxhir:
Description: This PR removes version 1.19.x from the NeoForge Documentation. This is because the version is not used and was clearly generated automatically with the default content for a new version.

This can be seen here: <https://docs.neoforged.net/docs/1.19.x/intro>

Note: I was not able to test the PR since I was having issues with npm and it not working when trying to start the local server.

------------------
Preview URL: https://pr-44.neoforged-docs-previews.pages.dev


ChampionAsh5357:
You need to remove the 1.19 mentions in the configurations as well.

Imzxhir:
I was able to test this PR and can confirm that it removes the 1.19.x version completely.

ChampionAsh5357 (approved):
Looks good.

PR #45, opened 2024-01-14
Update documentation on custom Advancement criteria

legobmw99:
A lot of the vanilla code for registering new Advancement criteria changed (I'm presuming in 1.20, but I hadn't much used it earlier). This includes both the names of classes and pretty large shifts from the currently documented requirements, such as no longer dealing directly with JSON objects. 

I believe this also should check one box of #20

------------------
Preview URL: https://pr-45.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/resources/server/advancements.md
+++ b/docs/resources/server/advancements.md
@@ -1,12 +1,10 @@
-Advancements
-============
+# Advancements
 
 Advancements are tasks that can be achieved by the player which may advance the progress of the game. Advancements can trigger based on any action the player may be directly involved in.
 
 All advancement implementations within vanilla are data driven via JSON. This means that a mod is not necessary to create a new advancement, only a [data pack][datapack]. A full list on how to create and put these advancements within the mod's `resources` can be found on the [Minecraft Wiki][wiki]. Additionally, advancements can be [loaded conditionally and defaulted][conditional] depending on what information is present (mod loaded, item exists, etc.). As with other data driven features, advancements can be generated via [data generators][datagen].
 
-Advancement Criteria
---------------------
+## Advancement Criteria
 
 To unlock an advancement, the specified criteria must be met. Criteria are tracked through triggers which execute when a certain action is performed: killing an entity, changing an inventory, breading animals, etc. Any time an advancement is loaded into the game, the criteria defined are read and added as listeners to the trigger. Afterwards a trigger function is called (usually named `#trigger`) which checks all listeners as to whether the current state meets the conditions of the advancement criteria. The criteria listeners for the advancement are only removed once the advancement has been obtained by completing all requirements.
 
@@ -82,7 +80,7 @@ public boolean matches(ItemStack stack) {
 
 The `SimpleCriterionTrigger<T>` subclass is responsible for specifying a codec to [serialize] the trigger instance `T` and supplying a method to check trigger instances and run attached listeners on success.
 
-The later is done by defining a method to check all trigger instances and run the listeners if their condition is met. This method takes in the `ServerPlayer` and whatever other data defined by the matching method in the `SimpleCriterionTrigger.SimpleInstance` subclass. This method should internally call `SimpleCriterionTrigger#trigger` to properly handle checking all listeners. Most trigger instances call this method `#trigger`.
+The latter is done by defining a method to check all trigger instances and run the listeners if their condition is met. This method takes in the `ServerPlayer` and whatever other data defined by the matching method in the `SimpleCriterionTrigger.SimpleInstance` subclass. This method should internally call `SimpleCriterionTrigger#trigger` to properly handle checking all listeners. Most trigger instances call this method `#trigger`.
 
 ```java
 // This method is unique for each trigger and is as such not a method to override
@@ -98,25 +96,31 @@ Finally, instances must be registered on the `Registries.TRIGGER_TYPE` registry.
 
 ### Serialization
 
-A [codec] must be defined to serialize and deserialize the trigger instance. Vanilla typically creates this codec as a constant within the instance implementation.
+A [codec] must be defined to serialize and deserialize the trigger instance. Vanilla typically creates this codec as a constant within the instance implementation that is then returned by the trigger's `#codec` method.
+
 
 ```java
-// in ExampleTriggerInstance
-public static final Codec<ExampleTriggerInstance> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+class ExampleTrigger extends SimpleCriterionTrigger<ExampleTrigger.ExampleTriggerInstance> {
+  @Override
+  public Codec<ExampleTriggerInstance> codec() {
+    return ExampleTriggerInstance.CODEC;
+  }
+  // ...
+  public class ExampleTriggerInstance implements SimpleCriterionTrigger.SimpleInstance {
+    public static final Codec<ExampleTriggerInstance> CODEC = ...;
+    // ...
+  }
+}
+```
+
+For the earlier example of a record with a `ContextAwarePredicate` and an `ItemPredicate`, the codec could be:
+
+```java
+RecordCodecBuilder.create(instance -> instance.group(
   ExtraCodecs.strictOptionalField(EntityPredicate.ADVANCEMENT_CODEC, "player").forGetter(ExampleTriggerInstance::player),
   ItemPredicate.CODEC.fieldOf("item").forGetter(ExampleTriggerInstance::item)
 ).apply(instance, ExampleTriggerInstance::new));
-```
-
-This is then returned by the trigger's `#codec` method.
-
-```java
-// in ExampleTrigger
-@Override
-public Codec<ExampleTriggerInstance> codec() {
-  return ExampleTriggerInstance.CODEC;
-}
-```
+``````
 
 ### Calling the Trigger
 
@@ -131,8 +135,7 @@ public void performExampleAction(ServerPlayer player, ItemStack stack) {
 }
 ```
 
-Advancement Rewards
--------------------
+## Advancement Rewards
 
 When an advancement is completed, rewards may be given out. These can be a combination of experience points, loot tables, recipes for the recipe book, or a [function] executed as a creative player.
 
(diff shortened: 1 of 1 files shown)


docs/resources/server/advancements.md
```diff
@@ -43,41 +43,38 @@ A list of criteria triggers defined by vanilla can be found in `CriteriaTriggers
...
-  super(ID, player);
-  // Store the item condition that must be met
+public record ExampleTriggerInstance(Optional<ContextAwarePredicate> player, ItemPredicate item) implements SimpleCriterionTrigger.SimpleInstance {
+  // extra methods here
 }
 ```
 
 :::note
-Typically, trigger instances have a static constructor which allow these instances to be easily created for data generation. These static factory methods can also be statically imported instead of the class itself.
+Typically, trigger instances have static helper methods which construct the full `Criterion<T>` object from the arguments to the instance. This allows these instances to be easily created during data generation, but are optional.
```
ChampionAsh5357, line 61: I would also specify they have some sort of predicate function which can be used to validate whether the condition specified has been met.

docs/resources/server/advancements.md
```diff
@@ -43,41 +43,38 @@ A list of criteria triggers defined by vanilla can be found in `CriteriaTriggers
...
 
 ```java
-public static ExampleTriggerInstance instance(ContextAwarePredicate player, ItemPredicate item) {
-  return new ExampleTriggerInstance(player, item);
+public static Criterion<ExampleTriggerInstance> instance(ContextAwarePredicate player, ItemPredicate item) {
+  return EXAMPLE_TRIGGER.get().createCriterion(new ExampleTriggerInstance(Optional.of(player), item));
```
ChampionAsh5357, line 65: Comment mentioning `EXAMPLE_TRIGGER` is a `DeferredHolder<CriterionTrigger<?>>`

docs/resources/server/advancements.md
```diff
@@ -43,41 +43,38 @@ A list of criteria triggers defined by vanilla can be found in `CriteriaTriggers
...
 
 ```java
-@Override
-public JsonObject serializeToJson(SerializationContext context) {
-  JsonObject obj = super.serializeToJson(context);
-  // Write conditions to json
-  return obj;
-}
+public static final Codec<ExampleTriggerInstance> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+  ContextAwarePredicate.CODEC.optionalFieldOf("player").forGetter(ExampleTriggerInstance::player),
```
ChampionAsh5357, line 75: `Optional`s for trigger instances are usually wrapped via `ExtraCodecs#strictOptionalField` instead of using `#optionalFieldOf`. This is because the latter will always be mapped as a success, regardless of if parsing fails. Meanwhile, if the former fails to parse, then an error result will be provided instead.

docs/resources/server/advancements.md
```diff
@@ -43,41 +43,38 @@ A list of criteria triggers defined by vanilla can be found in `CriteriaTriggers
...
-  return new ExampleTriggerInstance(player, item);
+public static Criterion<ExampleTriggerInstance> instance(ContextAwarePredicate player, ItemPredicate item) {
+  return EXAMPLE_TRIGGER.get().createCriterion(new ExampleTriggerInstance(Optional.of(player), item));
 }
 ```
 :::
 
-Additionally, the `#serializeToJson` method should be overridden. The method should add the conditions of the instance to the other JSON data.
+While it could be defined elsewhere, it is also typical to define a [`Codec`][codec] which can serialize this class as a constant in the class.
+A codec which can serialize this class must be defined somewhere, as it is required to implement `SimpleCriterionTrigger#codec` and is used to serialize criteria to JSON during data generation and deserialize them during loading.
```
ChampionAsh5357, line 71:
```suggestion
A [codec] must be defined to serialize and deserialize the trigger instance. Vanilla typically creates this codec as a constant within the instance implementation. 
```

Making the wording more concise.

docs/resources/server/advancements.md
```diff
@@ -92,48 +89,42 @@ public boolean matches(ItemStack stack) {
 
 ### SimpleCriterionTrigger
 
-The `SimpleCriterionTrigger<T>` subclass, where `T` is the type of the trigger instance, is responsible for specifying the registry name of the trigger, creating a trigger instance, and a method to check trigger instances and run attached listeners on success.
-
-The registry name of the trigger is supplied to `#getId`. This should match the registry name supplied to the trigger instance.
+The `SimpleCriterionTrigger<T>` subclass, where `T` is the type of the trigger instance, is responsible for specifying a codec to serialize `T`s, and supplying a method to check trigger instances and run attached listeners on success.
 
-A trigger instance is created via `#createInstance`. This method reads a criteria from JSON.
+The codec is typically defined as a constant of the inner `SimpleCriterionTrigger.SimpleInstance` class, and simply returned by the trigger's `codec()` method.
```
ChampionAsh5357, line 94: Wasn't this mentioned above as well? Where the codec lives should probably be restructured in this doc.

docs/resources/server/advancements.md
```diff
@@ -92,48 +89,42 @@ public boolean matches(ItemStack stack) {
...
 
 ```java
+// in ExampleTrigger
 @Override
-public ExampleTriggerInstance createInstance(JsonObject json, ContextAwarePredicate player, DeserializationContext context) {
-  // Read conditions from JSON: item
-  return new ExampleTriggerInstance(player, item);
+public Codec<ExampleTriggerInstance> codec() {
+  return ExampleTriggerInstance.CODEC;
 }
```
ChampionAsh5357, line 101: It would be better to combine all of the codec code blocks together.

docs/resources/server/advancements.md
```diff
@@ -92,48 +89,42 @@ public boolean matches(ItemStack stack) {
...
 ```
 
-Finally, a method is defined to check all trigger instances and run the listeners if their condition is met. This method takes in the `ServerPlayer` and whatever other data defined by the matching method in the `AbstractCriterionTriggerInstance` subclass. This method should internally call `SimpleCriterionTrigger#trigger` to properly handle checking all listeners. Most trigger instances call this method `#trigger`.
+Finally, a method is defined to check all trigger instances and run the listeners if their condition is met. This method takes in the `ServerPlayer` and whatever other data defined by the matching method in the `SimpleCriterionTrigger.SimpleInstance` subclass. This method should internally call `SimpleCriterionTrigger#trigger` to properly handle checking all listeners. Most trigger instances call this method `#trigger`.
 
 ```java
-// This method is unique for each trigger and is as such not overridden
+// This method is unique for each trigger and is as such not an override
```
ChampionAsh5357, line 107:
```suggestion
// This method is unique for each trigger and is as such not a method to override
```

ChampionAsh5357 (changes requested):
Bunch of general and clarification comments. Also, the doc needs to be slightly restructured since it feels like there are a few repeating parts.

docs/resources/server/advancements.md
legobmw99: That appears a bit lower as the final item

docs/resources/server/advancements.md
legobmw99: I didn't understand the distinction myself, thanks

docs/resources/server/advancements.md
legobmw99: Gave that a go. It feels a bit awkward to retroactively describe things which go in both classes, but I agree it's better than the duplication

docs/resources/server/advancements.md
```diff
@@ -92,48 +80,54 @@ public boolean matches(ItemStack stack) {
...
-  // Read conditions from JSON: item
-  return new ExampleTriggerInstance(player, item);
-}
-```
+The `SimpleCriterionTrigger<T>` subclass is responsible for specifying a codec to [serialize] the trigger instance `T` and supplying a method to check trigger instances and run attached listeners on success.
 
-Finally, a method is defined to check all trigger instances and run the listeners if their condition is met. This method takes in the `ServerPlayer` and whatever other data defined by the matching method in the `AbstractCriterionTriggerInstance` subclass. This method should internally call `SimpleCriterionTrigger#trigger` to properly handle checking all listeners. Most trigger instances call this method `#trigger`.
+The later is done by defining a method to check all trigger instances and run the listeners if their condition is met. This method takes in the `ServerPlayer` and whatever other data defined by the matching method in the `SimpleCriterionTrigger.SimpleInstance` subclass. This method should internally call `SimpleCriterionTrigger#trigger` to properly handle checking all listeners. Most trigger instances call this method `#trigger`.
```
ChampionAsh5357, line 85:
```suggestion
The latter is done by defining a method to check all trigger instances and run the listeners if their condition is met. This method takes in the `ServerPlayer` and whatever other data defined by the matching method in the `SimpleCriterionTrigger.SimpleInstance` subclass. This method should internally call `SimpleCriterionTrigger#trigger` to properly handle checking all listeners. Most trigger instances call this method `#trigger`.
```

Spelling because it's simple.

docs/resources/server/advancements.md
```diff
@@ -92,48 +80,54 @@ public boolean matches(ItemStack stack) {
...
+```
+
+This is then returned by the trigger's `#codec` method.
+
+```java
+// in ExampleTrigger
+@Override
+public Codec<ExampleTriggerInstance> codec() {
+  return ExampleTriggerInstance.CODEC;
+}
```
ChampionAsh5357, line 118: I think just for ease of understanding, the structure should be something like this with the comments:

```java
class ExampleTrigger<Instance> {
    codec() { Instance.CODEC; }

    class Instance {
        CODEC = ...;
    }
}
```

ChampionAsh5357 (changes requested):
Two minor changes, but otherwise looks good enough.

IchHabeHunger54:
I'm gonna request changing all the headers from == and -- to # and ## while you're at it.

docs/resources/server/advancements.md
legobmw99: Agreed

ChampionAsh5357 (approved):

XFactHD (approved):

IchHabeHunger54 (approved):

PR #48, opened 2024-01-29
Rework the Concepts chapter

IchHabeHunger54:
Next up in the series of major rewrites: the Concepts chapter! This PR:

- Rewrites the Registries page. Among other things, the page now contains the contents of the registry rework, as well as information on datapack registries.
- Rewrites the Events page. It now contains the changes since the split from Forge (for example the `ICancellable` interface), and is just better to read in general.
- Rewrites and significantly simplifies the Sides page.
- Removes the old mini-Resources article and replaces it with a more detailed article on Resource Locations and Resource Keys in the Miscellaneous section.
- Merges the Lifecycle article into the Events page.
- Merges the Datapack Registry Datagen article into the Registries page.
- Adds information about `DeferredRegister.Blocks` and `DeferredRegister.Items` to the blocks and items articles, respectively.
- Changes a lot of links due to the restructuring.

Please proofread the new articles before approving. This goes especially for the registries part, since I'm not certain I got everything correct there.

------------------
Preview URL: https://pr-48.neoforged-docs-previews.pages.dev


docs/blocks/index.md
```diff
@@ -8,10 +8,10 @@ One Block to Rule Them All
 
 Before we get started, it is important to understand that there is only ever one of each block in the game. A world consists of thousands of references to that one block in different locations. In other words, the same block is just displayed a lot of times.
 
-Due to this, a block should only ever be instantiated once, and that is during registration. Once the block is registered, you can then use the registered reference as needed. Consider this example (see the [Registration][registration] page if you do not know what you are looking at):
+Due to this, a block should only ever be instantiated once, and that is during [registration]. Once the block is registered, you can then use the registered reference as needed. Consider this example:
 
 ```java
-//BLOCKS is a DeferredRegister.Blocks
```
Matyrobbrt, line 14: This is not correct, Blocks is a specialized DR subclass, please revert this change.

docs/concepts/events.md
```diff
@@ -1,148 +1,174 @@
...
-	// ...
+@Mod("yourmodid")
+public class YourMod {
+    public YourMod(IEventBus modBus) {
+        NeoForge.EVENT_BUS.addListener(YourMod::onLivingJump);
+    }
+
+    // Heals an entity by half a heart every time they jump.
+    private static void onLivingJump(LivingJumpEvent event) {
+        event.getEntity().heal(1);
```
Matyrobbrt, line 24: This should be behind a side check.

docs/concepts/events.md
```diff
@@ -1,148 +1,174 @@
...
 
 ```java
-public class MyForgeEventHandler {
+public class EventHandler {
 	@SubscribeEvent
-	public void pickupItem(EntityItemPickupEvent event) {
-		System.out.println("Item picked up!");
-	}
+    public void onLivingJump(LivingJumpEvent event) {
+        event.getEntity().heal(1);
```
Matyrobbrt, line 37: Same as above, shouldn't heal directly on the client.

docs/concepts/events.md
```diff
@@ -1,148 +1,174 @@
...
 
 ```java
-public class MyStaticForgeEventHandler {
+public class EventHandler {
 	@SubscribeEvent
-	public static void arrowNocked(ArrowNockEvent event) {
-		System.out.println("Arrow nocked!");
-	}
+    public static void onLivingJump(LivingJumpEvent event) {
+        event.getEntity().heal(1);
```
Matyrobbrt, line 55: Same as above.

docs/concepts/events.md
```diff
@@ -1,148 +1,174 @@
...
+public class YourMod {
+    public YourMod(IEventBus modBus) {
+        NeoForge.EVENT_BUS.addListener(EventHandler.class);
+    }
+}
+```
 
-### Automatically Registering Static Event Handlers
+:::info
+`@SubscribeEvent` annotated methods must not be `private`. Any out of `package-private`, `protected` and `public` will work.
```
Matyrobbrt, line 68: `private` should work now too.

docs/concepts/events.md
```diff
@@ -1,148 +1,174 @@
...
-public class MyStaticClientOnlyEventHandler {
-	@SubscribeEvent
-	public static void drawLast(RenderLevelStageEvent event) {
-		System.out.println("Drawing!");
-	}
+@Mod.EventBusSubscriber(modid = "yourmodid")
+public class EventHandler {
+    @SubscribeEvent
+    public static void onLivingJump(LivingJumpEvent event) {
+        event.getEntity().heal(1);
```
Matyrobbrt, line 82: Same as above.

docs/concepts/registries.md
```diff
@@ -1,200 +1,232 @@
...
+Be aware that a few places explicitly require a `DeferredHolder` and will not just accept any `Supplier`. If you need a `DeferredHolder`, it is best to change the type of your `Supplier` back to `DeferredHolder`.
 
-Registered objects must always be referenced through a `RegistryObject` or a field with `@ObjectHolder`.
+Finally, since the entire system is a wrapper around registry events, we need to tell the `DeferredRegister` to attach itself to the registry events as needed:
 
-### Using RegistryObjects
+```java
+//This is our mod constructor
+public ExampleMod(IModEventBus bus) {
+    ExampleBlocksClass.BLOCKS.register(bus);
```
Matyrobbrt, line 69: Highlight this line.

docs/concepts/registries.md
```diff
@@ -1,200 +1,232 @@
...
-  * _A compile-time exception is thrown if a field does not have a corresponding registry or name._
-* _An exception is thrown if the resulting `ResourceLocation` is incomplete or invalid (non-valid characters in path)_
-* If no other errors or exceptions occur, the field will be injected
-* If all of the above rules do not apply, no action will be taken (and a message may be logged)
+```java
+// We use spells as an example for the registry here, without any details about what a spell actually is (as it doesn't matter).
+// Of course, all mentions of spells can and should be replaced with whatever your registry actually is.
+public static final ResourceKey<Registry<Spell>> SPELL_REGISTRY_KEY = ResourceKey.createRegistryKey(new ResourceLocation("yourmodid", "spells"));
+public static final Registry<YourRegistryContents> SPELL_REGISTRY = new RegistryBuilder<>(SPELL_REGISTRY_KEY)
+        // If you want the registry to sync its values.
```
Matyrobbrt, line 110: This is misleading. The sync option syncs registry IDs and is also discouraged as you should usually send actual ids instead of int ids. Should also note that synced registries will prevent clients from connecting to servers if they're missing entries present on the other side.

docs/concepts/registries.md
```diff
@@ -1,200 +1,232 @@
...
+        .sync(true)
+        // The default key. Similar to minecraft:air for blocks.
```
Matyrobbrt, line 112: It's optional.

docs/concepts/registries.md
```diff
@@ -1,200 +1,232 @@
...
-:::caution
-`DeferredRegister#makeRegistry` must be called before the `DeferredRegister` is added to the mod event bus via `#register`. `#makeRegistry` also uses the `#register` method to create the registry during `NewRegistryEvent`.
-:::
-
-Handling Missing Entries
-------------------------
+### `JsonCodecProvider`
 
-There are cases where certain registry objects will cease to exist whenever a mod is updated or, more likely, removed. It is possible to specify actions to handle the missing mapping through the third of the registry events: `MissingMappingsEvent`. Within this event, a list of missing mappings can be obtained either by `#getMappings` given a registry key and mod id or all mappings via `#getAllMappings` given a registry key.
+NeoForge provides a data provider for datapack registry elements that, given a registry key and a map of objects to generate, generates all JSON files for the objects in the map.
```
Matyrobbrt, line 169: The correct way of generating dp entries is through the use of `Bootstrap` and the registry access CF. Check the neo test mod for damage types as a reference.

Matyrobbrt (changes requested):

docs/concepts/registries.md
Matyrobbrt:
```suggestion
// If you want the registry to sync its registry entry IDs for use in networking
```

docs/concepts/registries.md
Matyrobbrt: https://github.com/neoforged/NeoForge/blob/5e1e9fad1f16b73d716b5d51a6a653594150c987/tests/src/main/java/net/neoforged/neoforge/debug/damagesource/DamageTypeTests.java#L80-L89

Matyrobbrt: We already have docs for this actually: https://docs.neoforged.net/docs/datagen/server/datapackregistries  
Though they may also need a bit of updating, and it has a broken admonition.

Matyrobbrt (approved):

docs/blocks/index.md
```diff
@@ -80,14 +92,40 @@ Directly using `Block` only allows for very basic blocks. If you want to add fun
...
+
+We already discussed how to create a `DeferredRegister.Blocks` [above], as well as that it returns `DeferredBlock`s. Now, let's have a look at what other utilities the specialized `DeferredRegister` has to offer. Let's start with `#registerBlock`:
+
+```java
+public static final DeferredRegister.Blocks BLOCKS = DeferredRegister.createBlocks("yourmodid");
+
+public static final DeferredBlock<Block> EXAMPLE_BLOCK = BLOCKS.registerBlock(
+        "example_block",
+        Block::new, // The factory that the properties will be passed into.
+        new BlockBehaviour.Properties() // The properties to use.
```
ChampionAsh5357, line 105: This constructor is private, isn't it? It's `BlockBehaviour.Properties#of`

docs/concepts/events.md
```diff
@@ -1,148 +1,183 @@
-Events
-======
+# Events
 
-Forge uses an event bus that allows mods to intercept events from various Vanilla and mod behaviors.
+One of NeoForge's main features is the event system. Events are fired for various things that happen in the game. For example, there are events for when the player right clicks, when a player or another entity jumps, when blocks are rendered, when the game is loaded, etc. A modder can subscribe event handlers to each of these events, and then perform their desired behavior inside these event handlers.
 
-Example: An event can be used to perform an action when a Vanilla stick is right-clicked.
+Events are fired on their respective event bus. The most important bus is `NeoForge.EVENT_BUS`. Besides that, during startup, a mod bus is spawned for each loaded mod and passed into the mod's constructor; mod bus event handlers can run in parallel, dramatically increasing startup speed. See [below][modbus] for more information.
```
ChampionAsh5357, line 5:
```suggestion
Events are fired on their respective event bus. The most important bus is `NeoForge.EVENT_BUS`. Besides that, during startup, a mod bus is spawned for each loaded mod and passed into the mod's constructor. Some mod bus event handlers can run in parallel, dramatically increasing startup speed. See [below][modbus] for more information.
```

Otherwise, it sounds like all mod bus methods run in parallel, which I don't believe is the case.

docs/concepts/events.md
```diff
@@ -1,148 +1,183 @@
...
-```java
-// In the main mod class ExampleMod
+Alternatively, event handlers can be annotation-driven by creating an event handler method and annotating it with `@SubscribeEvent`. Then, you can pass an instance of the encompassing class to the event bus, registering all `@SubscribeEvent`-annotated event handlers of that instance:
 
-// This event is on the mod bus
-private void modEventHandler(RegisterEvent event) {
-	// Do things here
+```java
+public class EventHandler {
+	@SubscribeEvent
```
ChampionAsh5357, line 39: Tab -> Space

docs/concepts/events.md
```diff
@@ -1,148 +1,183 @@
...
 ```
 
-which must be registered like this: `NeoForge.EVENT_BUS.register(MyStaticForgeEventHandler.class)`.
+## Event Options
 
-### Automatically Registering Static Event Handlers
+### Fields and Methods
 
-A class may be annotated with the `@Mod$EventBusSubscriber` annotation. Such a class is automatically registered to `NeoForge#EVENT_BUS` when the `@Mod` class itself is constructed. This is essentially equivalent to adding `NeoForge.EVENT_BUS.register(AnnotatedClass.class);` at the end of the `@Mod` class's constructor.
+Probably the most obvious part. Most events contain context for the event handler to use, such as an entity causing the event or a level the event occurs in.
```
ChampionAsh5357, line 100: What is the most obvious part?

docs/concepts/events.md
```diff
@@ -1,148 +1,183 @@
...
-Different events may use results in different ways, refer to the event's JavaDoc before using the result.
+Results are deprecated and will be replaced by more specific per-event results soon.
 :::
 
-Priority
---------
+### Priority
 
-Event handler methods (marked with `@SubscribeEvent`) have a priority. You can set the priority of an event handler method by setting the `priority` value of the annotation. The priority can be any value of the `EventPriority` enum (`HIGHEST`, `HIGH`, `NORMAL`, `LOW`, and `LOWEST`). Event handlers with priority `HIGHEST` are executed first and from there in descending order until `LOWEST` events which are executed last.
+Event handlers can optionally get assigned a priority. The `EventPriority` enum contains five values: `HIGHEST`, `HIGH`, `NORMAL` (default), `LOW` and `LOWEST`. Events are executed from highest to lowest priority, with undefined order for two events of the same priority.
```
ChampionAsh5357, line 126: Has priority been implemented for parallel events yet? If not, we may want to clarify this.

docs/concepts/registries.md
```diff
@@ -1,200 +1,317 @@
...
-class Holder {
-  @ObjectHolder(registryName = "minecraft:enchantment", value = "minecraft:flame")
-  public static final Enchantment flame = null;     // Annotation present. [public static] is required. [final] is optional.
-                                                    // Registry name is explicitly defined: "minecraft:enchantment"
-                                                    // Resource location is explicitly defined: "minecraft:flame"
-                                                    // To inject: "minecraft:flame" from the [Enchantment] registry
+// We use spells as an example for the registry here, without any details about what a spell actually is (as it doesn't matter).
+// Of course, all mentions of spells can and should be replaced with whatever your registry actually is.
+public static final ResourceKey<Registry<Spell>> SPELL_REGISTRY_KEY = ResourceKey.createRegistryKey(new ResourceLocation("yourmodid", "spells"));
+public static final Registry<YourRegistryContents> SPELL_REGISTRY = new RegistryBuilder<>(SPELL_REGISTRY_KEY)
```
ChampionAsh5357, line 151: This can be simplified with `DeferredRegister#makeRegistry`.

docs/items/index.md
```diff
@@ -54,6 +54,53 @@ Directly using `Item` only allows for very basic items. If you want to add funct
...
+```
+
+Internally, this will simply call `ITEMS.register("example_item", () -> new Item(new Item.Properties()))` by applying the properties parameter to the provided item factory (which is commonly the constructor).
+
+If you want to use `Item::new`, you can leave out the factory entirely:
+
+```java
+public static final Supplier<Item> EXAMPLE_ITEM = ITEMS.registerItem(
+        "example_item",
+        new ItemBehaviour.Properties() // The properties to use.
```
ChampionAsh5357, line 78:
```suggestion
        new Item.Properties() // The properties to use.
```

`ItemBehaviour` is not a thing.

docs/items/index.md
```diff
@@ -54,6 +54,53 @@ Directly using `Item` only allows for very basic items. If you want to add funct
...
+Finally, there's also shortcuts for block items:
+
+```java
+public static final Supplier<BlockItem> EXAMPLE_BLOCK_ITEM = ITEMS.registerSimpleBlockItem("example_block", ExampleBlocksClass.EXAMPLE_BLOCK, new Item.Properties());
+// Variant that omits the properties parameter:
+public static final Supplier<BlockItem> EXAMPLE_BLOCK_ITEM = ITEMS.registerSimpleBlockItem("example_block", ExampleBlocksClass.EXAMPLE_BLOCK);
+// Variant that omits the name parameter, instead using the block's registry name:
+public static final Supplier<BlockItem> EXAMPLE_BLOCK_ITEM = ITEMS.registerSimpleBlockItem(ExampleBlocksClass.EXAMPLE_BLOCK, new Item.Properties());
+// Variant that omits both the name and the properties:
+public static final Supplier<BlockItem> EXAMPLE_BLOCK_ITEM = ITEMS.registerSimpleBlockItem(ExampleBlocksClass.EXAMPLE_BLOCK);
```
ChampionAsh5357, line 101: Although it doesn't matter for most JREs, its probably best to mention that the block registry class should be classloaded before the item registry class is loaded if they are separated.

ChampionAsh5357 (changes requested):
Fine, but there are a few oversights.

docs/concepts/events.md
IchHabeHunger54: Hence the "can run in parallel" (as opposed to "do run in parallel"), but will adjust to better wording.

ChampionAsh5357: I understood the sentiment; however, my brain interpreted it as all regardless.

docs/misc/resourcelocation.md
```diff
@@ -0,0 +1,46 @@
+# Resource Locations
+
+`ResourceLocation`s are one of the most important things in Minecraft. They are used as keys in [registries][registries], as identifiers for data or resource files, as references to models in code, and in a lot of other places. A `ResourceLocation` consists of two parts: a namespace and a path, separated by a `:`.
+
+The namespace denotes what mod or datapack the location refers to. For example, a mod with the mod id `examplemod` will use the `examplemod` namespace. Minecraft uses the `minecraft` namespace. Extra namespaces can be defined at will simply by creating a corresponding data folder, this is usually done by datapacks to keep their logic separate from the point where they integrate with vanilla.
```
XFactHD, line 5:
```suggestion
The namespace denotes what mod, resourcepack or datapack the location refers to. For example, a mod with the mod id `examplemod` will use the `examplemod` namespace. Minecraft uses the `minecraft` namespace. Extra namespaces can be defined at will simply by creating a corresponding data folder, this is usually done by datapacks to keep their logic separate from the point where they integrate with vanilla.
```

docs/misc/resourcelocation.md
```diff
@@ -0,0 +1,46 @@
...
+The namespace and path of a `ResourceLocation` can be retrieved using `ResourceLocation#getNamespace()` and `#getPath()`, respectively, and the combined form can be retrieved through `ResourceLocation#toString`.
+
+`ResourceLocation`s are immutable. All utility methods on `ResourceLocation`, such as `withPrefix` or `withSuffix`, return a new `ResourceLocation`.
+
+## Resolving `ResourceLocation`s
+
+Some places, for example registries, use `ResourceLocation`s directly. Some other places, however, will resolve the `ResourceLocation` as needed. For example:
+
+- `ResourceLocation`s are used as identifiers for GUI background. For example, the furnace GUI uses the resource location `minecraft:textures/gui/container/furnace.png`. This maps to the file `assets/minecraft/textures/gui/container/furnace.png` on disk. Note that the `.png` suffix is required in this resource location.
+- `ResourceLocation`s are used as identifiers for block models. For example, the block model of dirt uses the resource location `minecraft:models/block/dirt`. This maps to the file `assets/minecraft/models/block/dirt.json` on disk. Note that the `.json` suffix is NOT required in this resource location.
```
XFactHD, line 26:
```suggestion
- `ResourceLocation`s are used as identifiers for block models. For example, the block model of dirt uses the resource location `minecraft:block/dirt`. This maps to the file `assets/minecraft/models/block/dirt.json` on disk. Note that the `.json` suffix is NOT required in this resource location.
```

docs/misc/resourcelocation.md
```diff
@@ -0,0 +1,46 @@
...
+- `ResourceLocation`s are used as identifiers for recipes. For example, the iron block crafting recipe uses the resource location `minecraft:iron_block`. This maps to the file `data/minecraft/recipes/iron_block.json` on disk. Note that the `.json` suffix is not required here. Note as well that this resource location automatically maps into the `recipes` subfolder.
+
+Whether the `ResourceLocation` expects a file suffix, or what exactly the resource location resolves to, depends on the use case.
+
+## `ModelResourceLocation`s
+
+`ModelResourceLocation`s are a special kind of resource location that includes a third part, called the variant. Minecraft uses these mainly to differentiate between different variants of item models, where the different variants are used in different display contexts (for example with tridents, which have different models in first person, third person and inventories).
```
XFactHD, line 33:
```suggestion
`ModelResourceLocation`s are a special kind of resource location that includes a third part, called the variant. Minecraft uses these mainly to differentiate between different variants of models, where the different variants are used in different display contexts (for example with tridents, which have different models in first person, third person and inventories).
```
The variant is always either `inventory` for item models or the comma-delimited string of property-value pairs (empty for blocks with no blockstate properties). This should be mentioned below in addition to the `inventory` example.

docs/misc/resourcelocation.md
```diff
@@ -0,0 +1,46 @@
...
+
+`ModelResourceLocation` is a [client only][sides] class. This means that servers referencing this class will crash with a `NoClassDefError`.
+
+## `ResourceKey`s
+
+`ResourceKey`s combine a registry id with a registry name. An example would be a registry key with the registry id `minecraft:item` and the registry name `minecraft:diamond_sword`. Unlike a `ResourceLocation`, `ResourceKey`s actually refer to a unique element, thus being able to clearly identify an element. They are most commonly used in contexts where many different registries come in contact with one another. A common use case are datapacks, especially worldgen.
+
+A new `ResourceKey` can be created through the static method `ResourceKey#create(ResourceKey<? extends Registry<T>>, ResourceLocation)`. The second parameter here is the registry name, while the first parameter is what is known as a registry key. Registry keys are a special kind of `ResourceKey` whose registry is the root registry (i.e. the registry of all other registries). A registry key can be created via `ResourceKey#createRegistryKey(ResourceLocation)` with the desired registry's id.
```
XFactHD, line 43: It may be worth mentioning that `ResourceKey`s are interned which on one hand allows them to be compared by reference equality and on the other hand means that their instantiation is comparatively expensive.

XFactHD (changes requested):

docs/concepts/registries.md
```diff
@@ -1,200 +1,317 @@
...
 ```
 
-### Registries that aren't Forge Registries
-
-Not all registries are wrapped by Forge. These can be static registries, like `LootItemConditionType`, which are safe to use. There are also dynamic registries, like `ConfiguredFeature` and some other worldgen registries, which are typically represented in JSON. `DeferredRegister#create` has an overload which allows modders to specify the registry key of which vanilla registry to create a `RegistryObject` for. The registry method and attaching to the mod event bus is the same as other `DeferredRegister`s.
+The class `DeferredHolder<R, T extends R>` holds our object. The type parameter `R` is the type of the registry we are registering to (in our case `Block`). The type parameter `T` is the type of our supplier. Since we directly register a `Block` in this example, we provide `Block` as the second parameter. If we were to register an object of a subclass of `Block`, for example `SlabBlock`, we would provide `SlabBlock` here instead.
 
-:::danger
-Dynamic registry objects can **only** be registered through data files (e.g. JSON). They **cannot** be registered in-code.
+:::note
```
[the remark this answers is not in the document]
IchHabeHunger54: It's not really error-prone though? As long as all the classes are classloaded during registration, there's usually no problem. (Unless you are referring to something different, in which case, please elaborate)

docs/concepts/registries.md
```diff
@@ -1,200 +1,317 @@
...
+Datapack registries can be obtained from a `RegistryAccess`. This `RegistryAccess` can be retrieved by calling `ServerLevel#registryAccess()` if on the server, or `Minecraft.getInstance().connection#registryAccess()` if on the client (the latter only works if you are actually connected to a world, as otherwise the connection will be null). The result of these calls can then be used like any other registry to get specific elements, or to iterate over the contents.
 
-When using `NewRegistryEvent`, calling `#create` with a `RegistryBuilder` will return a supplier-wrapped registry. The supplied registry can be accessed after `NewRegistryEvent` has finished posting to the mod event bus. Getting the custom registry from the supplier before `NewRegistryEvent` finishes firing will result in a `null` value.
+### Custom Datapack Registries
 
-#### New Datapack Registries
+Custom datapack registries are created through `RegistryBuilder` like all other registries, but are registered to `DataPackRegistryEvent.NewRegistry` instead of `NewRegistryEvent`. Reiterating the spells example from before, registering our spell registry as a datapack registry looks something like this:
```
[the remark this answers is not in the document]
IchHabeHunger54: Ah, I was mixing regular with DP registries. Adjusting.

docs/concepts/registries.md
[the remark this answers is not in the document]
XFactHD: It's error-prone in that people constantly forget to classload the class holding the registry entries at an adequate time and then get a crash when they access one of their objects during late startup or gameplay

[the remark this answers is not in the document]
IchHabeHunger54: Alright, gonna remove that recommendation.

docs/concepts/registries.md
```diff
@@ -1,200 +1,316 @@
...
+    });
+```
 
-| Action | Description |
-| :---:  |     :---    |
-| IGNORE | Ignores the missing entry and abandons the mapping. |
-|  WARN  | Generates a warning in the log. |
-|  FAIL  | Prevents the world from loading. |
-| REMAP  | Remaps the entry to an already registered, non-null object. |
+The `BootstrapContext` can also be used to lookup entries from another registry if needed:
```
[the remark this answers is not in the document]
sciwhiz12: Very scremworthy. Some good news at least is that they renamed the class in 24w04a to fix that typo.

To add to this: this should probably come with a note of some kind, to tell the reader/editor that this _isn't_ a typo in our docs and it is the actual name of the class (until 24w04a).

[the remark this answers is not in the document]
IchHabeHunger54: Honestly, given that vanilla fixed this in the 1.20.5 snapshots, I'm tempted to just leave it there.

ChampionAsh5357 (dismissed):

XFactHD (approved):
:shipit:

ChampionAsh5357 (approved):
Approving again

XFactHD (approved):

PR #52, opened 2024-02-19
Updated code snippets to use NeoForgeRegistries instead of ForgeRegis…

Auxy6858:
…tries


Matyrobbrt:
That entire page needs to be rewritten to account for the registry rework.  
Duplicate of #48

PR #60, opened 2024-03-13
Change the outdated emphasis block into the current theme

srcres258:
The "!!! important" usage within the documentation is used within the Forge epoch, however NeoForge turned to use a novel format for these highlighted paragraphs. Hence make up for the missed one.

------------------
Preview URL: https://pr-60.neoforged-docs-previews.pages.dev


ChampionAsh5357:
Why not do it for all the pages which have the issue?

srcres258:
> Why not do it for all the pages which have the issue?

Only found this one by accident when reading the documentation. If needed I can check all pages to discover similar issues.

ChampionAsh5357:
Please do and thanks. There's only a few more instances based on a quick search.

srcres258:
> Please do and thanks. There's only a few more instances based on a quick search.

OK, I'll work on then.

srcres258:
All documentation files have been checked and fixed up targeting the same issue.

PR #62, opened 2024-03-26
Rework the Client Resources chapter

IchHabeHunger54:
Probably the biggest rework yet: the Resources chapter. I originally wanted to redo the entire Resources chapter at once, but given the sheer scope of that, I decided to split it into two: client and server. So client resources rework first it is.

This PR:

- Reworks the old Resources article, explaining the similarities and differences between assets/data and resource/data packs, as well as giving a brief overview about what datagen is.
- Adds a new article about I18n and L10n. This article replaces the I18n article in the Concepts section, the Component article in the Misc section, and the Language Datagen article.
- Adds new articles about particles and sounds. These articles replaces the Particles and Sounds articles in the Game Effects section, as well as the Sound Definition Datagen article.
- Adds a new article about textures and texture metadata.
- Adds several new articles in relation to models. These are rewritten, condensed and updated versions of the various mini-articles in the Rendering folder, the old articles in the Models folder, and the old Model Datagen article. Information about several previously undocumented features, such as baked model wrappers, was added as well.
- Does a fair bit of restructuring. The Rendering and Game Effects folders are gone, as their contents were merged into the Client Resources folder. The client datagen folder is gone as well; as a result, the server datagen was moved to the root of the datagen folder.
- Changes the Contribution Guidelines to use `json5` for JSON blocks instead of `js` (to allow comments), [as discussed on Discord](https://discord.com/channels/313125603924639766/1185208532014997524/1216552938260926526). This guideline change is not applied to existing articles, only to new ones.
- Deletes a few now-unused pictures.
- Changes links as needed.

I made a few general design decisions in this PR that I'd like to outline here:

- Merge datagen articles into their respective "base articles" (lang datagen -> lang article, sounds datagen -> sounds article, etc.)
- Merge articles together. Instead of the previous 12 articles about model stuff, we have 4 now. These contain more content and are significantly longer, notably because I wanted to avoid single-page articles like the Tinting article. This philosophy can (and I imagine, will) be discussed, and I'm open to suggestions on what to change here.
- Use more examples. If there was one thing that was bad about the old docs, it was the lack of examples to work with. Just mentioning the methods is not a bad idea per se, but I see nothing wrong with also providing copy-paste base structures to work off. You need to insert your own names and stuff anyway. However, I do recognize that this contributes a lot to the length of some articles.

------------------
Preview URL: https://pr-62.neoforged-docs-previews.pages.dev


docs/resources/client/models/index.md
```diff
@@ -1,25 +1,291 @@
...
+
+```java
+@SubscribeEvent
+public static void registerBlockColorHandlers(RegisterColorHandlersEvent.Block event) {
+    // Parameters are the block's state, the level the block is in, the block's position, and the tint index.
+    event.register((state, level, pos, tintIndex) -> {
+            // Replace with your own calculation. See the BlockColors class for vanilla references.
+            // Generally, if the tint index is -1, it means that no tinting should take place
+            // and a default value should be used instead.
+            return 0xFFFFFF;
```
ChampionAsh5357, line 238: In RGB format.

docs/resources/client/models/modelloaders.md
```diff
@@ -0,0 +1,444 @@
...
+
+The entry point for a block model remains the model JSON file. However, you can specify a `loader` field in the root of the JSON that will swap out the default loader for your own loader. A custom model loader may ignore all fields the default loader requires.
+
+## Builtin Model Loaders
+
+Besides the default model loader, NeoForge offers a total of seven builtin loaders, each serving a different purpose.
+
+### Composite Model
+
+A composite model can be used to specify different model parts in the parent and only apply some of them in a child. This is best illustrated by an example. Consider the following parent model at `examplemod:example_composite_model`:
```
ChampionAsh5357, line 13: Composite models also can have different render types for each part, correct? Also, render order if I'm not mistaken.

docs/resources/client/models/modelloaders.md
```diff
@@ -0,0 +1,444 @@
...
+    "part_2": true
+  }
+}
+```
+
+To [datagen][modeldatagen] this model, use the custom loader class `CompositeModelBuilder`.
+
+### Dynamic Fluid Container Model
+
+The dynamic fluid container model, also called dynamic bucket model after its most common use case, is used for items that represent a fluid container (such as a bucket or a tank) and want to show the fluid within the model. This only works if there is a fixed amount of fluids (e.g. only lava and powder snow) that can be used, use a [block entity renderer][ber] instead if the fluid is arbitrary.
```
ChampionAsh5357, line 51: Doesn't this still have the base and cover which defines the container for the fluid model?

docs/resources/client/particles.md
```diff
@@ -0,0 +1,254 @@
...
+public static void registerParticleProviders(RegisterParticleProvidersEvent event) {
+    // There are multiple ways to register providers, all differing in the functional type they provide in the
+    // second parameter. For example, #registerSpriteSet represents a Function<SpriteSet, ParticleProvider<?>>:
+    event.registerSpriteSet(MyParticleTypes.MY_PARTICLE.get(), MyParticleProvider::new);
+    // Other methods include #registerSprite, which is essentially a Supplier<TextureSheetParticle>,
+    // and #registerSpecial, which maps to a Supplier<Particle>. See the source code of the event for further info.
+}
+```
+
+### Particle Definitions
```
ChampionAsh5357, line 102: This only applies if the sprite set method is used, when the `ParticleRenderType` uses the particles texture atlas.

docs/resources/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+Both resource and data packs normally require a [`pack.mcmeta` file][packmcmeta], this was also the case in past Forge versions. However, NeoForge generates these at runtime for you, so you don't need to worry about it anymore.
+
+If you are confused about the format of something, an easy way to change that is to have a look at the vanilla resources. Your NeoForge development environment not only contains vanilla code, but also vanilla resources. They can be found in the External Resources section (IntelliJ)/Project Libraries section (Eclipse), under the name `ng_dummy_ng.net.minecraft:client:client-extra:<version>` (for Minecraft resources) or `ng_dummy_ng.net.neoforged:neoforge:<version>` (for NeoForge resources).
```
ChampionAsh5357, line 9: Probably should specify `<version>` to be either `<minecraft_version>` or `<neoforge_version>`

docs/resources/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+NeoForge automatically applies all mod data packs to a new world upon creation. It is currently not possible to disable mod data packs. However, most data files can be overridden (and thus be removed by replacing them with an empty file) by a data pack with a higher priority. Additional data packs can be enabled or disabled by placing them in a world's `datapacks` subfolder and then enabling or disabling them through the [`/datapack`][datapackcmd] command.
+
+:::info
+There is currently no built-in way to apply a set of custom data packs to every world. However, there are a number of mods that achieve this.
+:::
+
+Data packs may contain folders with files affecting the following things (TODO add links):
```
ChampionAsh5357, line 33: TODO

docs/resources/index.md
```diff
@@ -0,0 +1,185 @@
...
+
+- `event.getGenerator()` returns the `DataGenerator` that you register the providers to.
+- `event.getPackOutput()` returns a `PackOutput` that is used by some providers to determine their file output location.
+- `event.getExistingFileHelper()` returns an `ExistingFileHelper` that is used by providers for things that can reference other files (for example block models, which can specify a parent file).
+- `event.getLookupProvider()` returns a `CompletableFuture<HolderLookup.Provider>` that is mainly used by tags and datagen registries to reference other, potentially not yet existing elements.
+- `event.includeClient()`, `event.includeServer()`, `event.includeDev()` and `event.includeReports()` are `boolean` methods that allow you to check whether specific command line arguments (see below) are enabled.
+
+### Command Line Arguments
+
+The data generator can accept several command line arguments:
```
ChampionAsh5357, line 134: Might be best to create a groovy example for the default cmd line arguments.

ChampionAsh5357 (changes requested):
Very nice, here are my overall comments.

docs/resources/client/models/bakedmodel.md
```diff
@@ -0,0 +1,117 @@
...
+| `boolean isCustomRenderer()`                                                  | If true, skips normal rendering and calls an associated [block entity renderer][ber]'s `renderByItem` method instead. If false, renders through the default renderer.                                          |
+| `ItemOverrides getOverrides()`                                                | Returns the [`ItemOverrides`][itemoverrides] associated with this model. This is generally only relevant on item models.                                                                                       |
+| `ModelData getModelData(BlockAndTintGetter, BlockPos, BlockState, ModelData)` | Returns the model data to use for the model. Mainly used for [block entity renderers][ber].                                                                                                                    |
+| `TextureAtlasSprite getParticleIcon(ModelData)`                               | Returns the particle sprite to use for the model. May use the model data to use different particle sprites for different model data values. NeoForge-added, replacing the vanilla overload with no parameters. |
```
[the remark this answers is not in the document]
IchHabeHunger54: For which method exactly?

docs/resources/client/models/modelloaders.md
IchHabeHunger54: I did a quick glance through the code and found nothing like that? You can definitely specify render types per part, yes, as they're simply models.

docs/resources/client/particles.md
IchHabeHunger54: Mind elaborating?

docs/resources/client/models/bakedmodel.md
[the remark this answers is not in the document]
ChampionAsh5357: getParticleTexture I think is the vanilla name.

docs/resources/client/models/modelloaders.md
ChampionAsh5357: Tbh, I don't remember where I was going with this, so you can resolve it.

docs/resources/client/particles.md
ChampionAsh5357: Particle descriptions are essentially a list of textures to be stitched into the particle texture atlas. When registering the particle factory, you have to use the sprite particle registration interface to tell the engine that you have textures to be stitched. An error is thrown if there is a description but incorrect factory registry, or vice versa. As the descriptions are stitched into the atlas, only particle render types that use that atlas will actually render the particle correctly.

docs/resources/client/models/index.md
```diff
@@ -1,25 +1,295 @@
...
```
ChampionAsh5357, line 242: Probably should be mentioned that vanilla implementations force the alpha to be 1f or 255.

ChampionAsh5357 (changes requested):
One more thought and the remaining unresolved comments. Otherwise, looks good

docs/resources/client/models/bakedmodel.md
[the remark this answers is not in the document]
IchHabeHunger54: Ah, that would be the `getParticleIcon()` overload without parameters that is already mentioned. Will add the name again for clarification.

docs/resources/client/models/bakedmodel.md
```diff
@@ -0,0 +1,117 @@
...
+
+- A `BlockState`: The [blockstate] being rendered. May be null, indicating that an item is being rendered.
+- A `Direction`: The direction of the face being culled against. May be null, indicating that no culling should occur. Will always be null for items.
+- A `RandomSource`: A client-bound random source you can use for randomization.
+- A `ModelData`: The extra model data to use. This may contain additional data from the block entity needed for rendering.
+- A `RenderType`: The [render type][rendertype] to use for rendering the block. May be null, indicating that the default render type should be used. The default render type is looked up in `ItemBlockRenderTypes`, and if no render type is found there either, `minecraft:solid` is used. Do never call on `ItemBlockRenderTypes` yourself!
```
embeddedt, line 15:
```suggestion
- A `RenderType`: The [render type][rendertype] to use for rendering the block. May be null, indicating that the default render type should be used. The default render type is looked up in `ItemBlockRenderTypes`, and if no render type is found there either, `minecraft:solid` is used. You should never call on `ItemBlockRenderTypes` yourself!
```

docs/resources/client/models/bakedmodel.md
```diff
@@ -0,0 +1,117 @@
...
+`BakedModel`s are the in-code representation of a shape with textures. They can be obtained from multiple sources, for example by calling `UnbakedModel#bake` (default model loader) or `IUnbakedGeometry#bake` ([custom model loaders][modelloader]). Some [block entity renderers][ber] also make use of baked models. There is no limit to how complex a model may be.
+
+## Methods of `BakedModel`
+
+### `getQuads`
+
+The most important method of a baked model is `getQuads`. This method is responsible for returning a list of `BakedQuad`s, which can then be sent to the GPU. A quad compares to a triangle in a modeling program (and in most other games), however due to Minecraft's general focus on squares, the developers elected to use quads (4 vertices) instead of triangles (3 vertices) for rendering in Minecraft. `getQuads` has five parameters that can be used:
+
+- A `BlockState`: The [blockstate] being rendered. May be null, indicating that an item is being rendered.
+- A `Direction`: The direction of the face being culled against. May be null, indicating that no culling should occur. Will always be null for items.
```
embeddedt, line 12: Items go through all possible directions (and null) just like blocks (see `ItemRenderer#renderModelLists`), so the direction is not always null for items.

docs/resources/client/models/bakedmodel.md
```diff
@@ -0,0 +1,117 @@
...
+- A `RandomSource`: A client-bound random source you can use for randomization.
+- A `ModelData`: The extra model data to use. This may contain additional data from the block entity needed for rendering.
```
embeddedt, line 14: The additional data can also be supplied by `BakedModel#getModelData` (see my comment there). Also, might be worth mentioning this is basically unused and always `EMPTY` for items.

docs/resources/client/models/bakedmodel.md
```diff
@@ -0,0 +1,117 @@
...
+| `boolean isCustomRenderer()`                                                  | If true, skips normal rendering and calls an associated [block entity renderer][ber]'s `renderByItem` method instead. If false, renders through the default renderer.                                                              |
+| `ItemOverrides getOverrides()`                                                | Returns the [`ItemOverrides`][itemoverrides] associated with this model. This is generally only relevant on item models.                                                                                                           |
+| `ModelData getModelData(BlockAndTintGetter, BlockPos, BlockState, ModelData)` | Returns the model data to use for the model. Mainly used for [block entity renderers][ber].                                                                                                                                        |
```
embeddedt, line 38: Actually `getModelData` is probably used more often by models for blocks without a block entity (e.g. connected texture mods, since they can figure out what blocks the current block is surrounded by). Block entities already have a way of supplying model data via `BlockEntity#getModelData`.

docs/resources/client/models/bakedmodel.md
```diff
@@ -0,0 +1,117 @@
...
+    // Pass the original model to super.
+    public MyBakedModelWrapper(BakedModel originalModel) {
+        super(originalModel);
+    }
+    
+    // Override whatever methods you want here. You may also access originalModel if needed.
+}
+```
+
+After writing your model wrapper class, you must apply the wrappers to the models it should affect. Do so in a [client-side][sides] [event handler][event] for `ModelEvent.ModifyBakingResult`:
```
embeddedt, line 86: I'm not sure we should be encouraging wrapping this prominently in the documentation. I have seen several mods abuse it when they should just use a custom model loader for their models instead.

embeddedt (changes requested):
I only looked at the baked model stuff I am familiar with.

docs/resources/client/models/bakedmodel.md
ChampionAsh5357: Then it's probably better to list when it makes to use a wrapper or a custom model loader.

docs/resources/client/sounds.md
```diff
@@ -0,0 +1,312 @@
...
+
+- `playSound(SoundEvent soundEvent, float volume, float pitch)`
+  - Forwards to `Level#playSound` with `null` as the player, `SoundSource.ENTITY` as the sound source, the entity's position for x/y/z, and the other parameters passed in.
+
+### `Player`
+
+- `playSound(SoundEvent soundEvent, float volume, float pitch)` (overrides the method in `Entity`)
+  - Forwards to `Level#playSound` with `this` as the player, `SoundSource.PLAYER` as the sound source, the player's position for x/y/z, and the other parameters passed in. As such, the client/server behavior mimics the one from `Level#playSound`:
+    - Client behavior: Play the sound event to the client player at the given location.
+    - Server behavior: Play the sound event to everyone near the given location except the client player.
```
XFactHD, line 238:
```suggestion
    - Server behavior: Play the sound event to everyone near the given location except the player this method was called on.
```

docs/resources/client/textures.md
```diff
@@ -0,0 +1,44 @@
...
+
+Textures should generally be in sizes that are powers of two, for example 16x16 or 32x32. Unlike older versions, modern Minecraft natively supports block and item texture sizes greater than 16x16. For textures that are not in powers of two that you render yourself anyway (for example GUI backgrounds), create an empty file in the next available power-of-two size (often 256x256), and add your texture in the top left corner of that file, leaving the rest of the file empty. The actual size of the drawn texture can then be set in the code that uses the texture.
+
+## Texture Metadata
+
+Texture metadata can be specified in a file named exactly the same as the texture, with an additional `.mcmeta` suffix. For example, an animated texture at `textures/block/example.png` would need an accompanying `textures/block/example.png.mcmeta` file. The `.mcmeta` file has the following format:
+
+```json5
+{
+  // Whether the texture will be blurred if needed. Defaults to false. Currently unused (but functional).
```
XFactHD, line 13: Where are these two loaded such that they would be functional?

docs/resources/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+Both resource and data packs normally require a [`pack.mcmeta` file][packmcmeta], this was also the case in past Forge versions. However, NeoForge generates these at runtime for you, so you don't need to worry about it anymore.
+
+If you are confused about the format of something, an easy way to change that is to have a look at the vanilla resources. Your NeoForge development environment not only contains vanilla code, but also vanilla resources. They can be found in the External Resources section (IntelliJ)/Project Libraries section (Eclipse), under the name `ng_dummy_ng.net.minecraft:client:client-extra:<minecraft_version>` (for Minecraft resources) or `ng_dummy_ng.net.neoforged:neoforge:<neoforge_version>` (for NeoForge resources).
```
XFactHD, line 9: > an easy way to change that is to have a look [...]

This sounds really weird, that should be worded differently

docs/resources/index.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Data Generation
+
+Data generation, colloquially known as datagen, is a way to programmatically generate JSON resource files, in order to avoid the tedious and error-prone process of writing them by hand. The name is a bit misleading, as it works for assets as well as data.
+
+Datagen is run through the Data run configuration, which is generated for you alongside the Client and Server run configurations. The data run configuration follows the [mod lifecycle][lifecycle] until after the registry events are fired. It then fires the [`GatherDataEvent`][event], in which you can register your to-be-generated objects in the form of data providers, writes said objects to disk, and ends the process.
+
+All data providers extend the `DataProvider` interface and usually require one method to be overridden. The following is a list of all data generators Minecraft and NeoForge offer (the linked articles add further information, such as helper methods):
```
XFactHD, line 67: This is by far not all data providers offered by Minecraft and NeoForge, it is missing all the "report providers" (though those are rarely relevant to mods) and the `SpriteSourceProvider` and its parent, the `JsonCodecProvider`, and likely some others I missed. This should be worded more along the lines of "The following is a list of some noteworthy data generators Minecraft and NeoForge offer"

XFactHD (changes requested):

docs/resources/client/models/bakedmodel.md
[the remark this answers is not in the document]
IchHabeHunger54: Would you mind elaborating on this?

docs/resources/client/models/bakedmodel.md
```diff
@@ -43,6 +43,8 @@ Other methods in `BakedModel` that you may override and/or query include:
...
 | `TextureAtlasSprite getParticleIcon(ModelData)`                               | Returns the particle sprite to use for the model. May use the model data to use different particle sprites for different model data values. NeoForge-added, replacing the vanilla `getParticleIcon()` overload with no parameters.                                                                                                                                                                          |
+| `ChunkRenderTypeSet getRenderTypes(BlockState, RandomSource, ModelData)`      | Returns a `ChunkRenderTypeSet` containing the render type(s) to use for rendering the block model. By default falls back to the normal model-bound render type lookup, which always yields a set with one element. Note that `ChunkRenderTypeSet` is not actually a set, but an ordered `Iterable<List>` (so more of a list). Only used for block models, item models use the overload below.               |
```
XFactHD, line 46: > By default falls back to the normal model-bound render type lookup

What is "normal model-bound render type lookup" referring to? This should clarify that certain vanilla models are patched to use the render type specified in the model JSON if present and otherwise fall back to an `ItemBlockRenderTypes` lookup with the given `BlockState`. This also applies to the item overload below, except that the lookup is done with the given `ItemStack`.

> Note that `ChunkRenderTypeSet` is not actually a set, but an ordered `Iterable<List>`

Apart from implementation details being completely irrelevant here, the `ChunkRenderTypeSet` is actually a set in that it is backed by a `BitSet` and can only contain each `RenderType` exactly once.

docs/resources/client/models/bakedmodel.md
```diff
@@ -0,0 +1,132 @@
...
+        // The model resource location of the model to modify.
+        new ModelResourceLocation("examplemod", "example_item", "inventory"),
+        // A BiFunction with the location and the original models as parameters, returning the new model.
+        (location, model) -> new MyBakedModelWrapper(model);
+    );
+}
+```
+
+:::warning
+It is generally encouraged to use a [custom model loader][modelloader] over baked model wrappers when possible. Custom model loaders can also use `BakedModelWrapper`s if needed.
```
XFactHD, line 118:
```suggestion
It is generally encouraged to use a [custom model loader][modelloader] over wrapping baked models in `ModelEvent.ModifyBakingResult` when possible. Custom model loaders can also use `BakedModelWrapper`s if needed.
```

XFactHD (changes requested):

XFactHD (approved):
:shipit:

ChampionAsh5357 (approved):
Looks fine.

embeddedt (approved):

PR #70, opened 2024-05-13
Overhaul the mod files article

### LINKED ISSUE #63: Clarifications needed to 'change the name of a mod' in "Getting Started with NeoForge"
So, I'm starting in minecraft modding, and found it very unclear when trying to change the name of my mod. As it reads on the document, on the section "Customizing Your Mod Information"

_"Many of the **basic properties** of your mod **can be changed in the gradle.properties file**. This includes basic things like the **mod name** or the mod version."_

Further by investigating, you stumble across mod files and find the definition of the property "mod_id"

_"The id of your mod. This should be something unique and memorable, as having two mods with the same id will prevent the game from loading. The mod id shows up in a lot of places, for example as the namespace for all your registered things, or as the namespace for your resource and data packs."_

One starting would think that in order to change the mod name you would only need to change this variable, but it turns out that there are more elements that interfere when building the mod and are related to the name, so if you only changed that property, it would cause trouble for mods to properly build.

Might be redundant for some but wouldn't it be more clear to say or suggest in the for example that in order to change the name of the mod, there are other properties that need to be changed properly?
For example, in the definition of "mod_id":

_"Note that this is not the only property related to the mod name, only changing this variable can create issues"._

Regardless of that issue, thanks in advance for the support.

IchHabeHunger54:
Fixes and improves on various minor issues with the mod files article, including better explanations of how mod IDs work, removing pack format stuff, removing the mention of mod name translations (which were removed from Neo), and table and italics reformatting. Fixes #63.

------------------
Preview URL: https://pr-70.neoforged-docs-previews.pages.dev


docs/gettingstarted/modfiles.md
```diff
@@ -2,30 +2,40 @@
...
+| `mod_authors`             | The authors of the mod, shown in the mod list.                                                                                                                                                                                                        | `mod_authors=ExampleModder`               |
+| `mod_description`         | The description of the mod, as a multiline string, shown in the mod list. Newline characters (`\n`) can be used and will be replaced properly.                                                                                                        | `mod_authors=Example mod description.`    |
+| `pack_format_number`      | The version number of your mod's data and resource pack. Mojang bumps this without something that could be considered a consistent scheme, so it's best to just look up what the current number is. As of Minecraft 1.20.2, the pack version is `18`. | `pack_version_number=18`                  |
```
ChampionAsh5357, line 28: Honestly, we should remove this as the pack format can be datagenned and that was the recommended way until NeoForge removed the requirement.

Actually, on another look, this has been removed from the MDK, so this can be removed here.

docs/gettingstarted/modfiles.md
```diff
@@ -2,30 +2,40 @@
...
+
+### The Mod ID
+
+The mod ID is the main way your mod is distinguished from others. It is used in a wide variety of places, including as the namespace for your mod's [registries][registration], and as your [resource and data pack][resource] namespaces. Having two mods with the same id will prevent the game from loading.
+
+As such, your mod ID should be something unique and memorable. Usually, it will be your mod's display name (but lower case), or some variation thereof. Mod IDs may only contain lowercase letters, digits and underscores.
```
ChampionAsh5357, line 34: Probably should specify the minimum and maximum character limit.

ChampionAsh5357 (changes requested):
Two minor things I noticed while double-checking everything.

ChampionAsh5357 (changes requested):

docs/gettingstarted/modfiles.md
```diff
@@ -91,27 +100,32 @@ modId = "examplemod1"
...
+| `version`       | string   | `"1"`                        | The version of the mod, preferably in a [variation of Maven versioning][versioning]. When set to `${file.jarVersion}`, it will be replaced with the value of the `Implementation-Version` property in the JAR's manifest (displays as `0.0NONE` in a development environment). | `version="1.20.2-1.0.0"`                                        |
+| `displayName`   | string   | value of `modId`             | The display name of the mod. Used when representing the mod on a screen (e.g., mod list, mod mismatch). This value is also translatable, see [Translating Mod Metadata][i18n] for more info.                                                                                   | `displayName="Example Mod"`                                     |
```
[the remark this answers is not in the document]
IchHabeHunger54: Oh damnit, I added that to the wrong line.

ChampionAsh5357 (approved):
Looks good now. Will merge in a bit.

PR #73, opened 2024-05-15
Update README with correct info and add .nvmrc

ChampionAsh5357:
Adds an `.nvmrc` file to be used by `nvm` and updates the README with the new information.

Also fixes a small typo where we migrated to Docusaurus 3.

------------------
Preview URL: https://pr-73.neoforged-docs-previews.pages.dev


README.md
```diff
@@ -2,14 +2,22 @@
...
 ## Contributing
 
 You can read the [contribution guidelines on the docs](https://docs.neoforged.net/contributing/).
 
 If you wish to contribute to the documentation, fork and clone this repository.
 
+The documentation uses Node.js 18. This can either be install manually or using a version manager that supports `.node-version` or `.nvmrc`. One or both of the following commands can be used to install and use the correct Node.js version:
+
+```bash
+# Replace [version_manager] with your node version manager (e.g., nvm, nvs, etc.)
```
IchHabeHunger54, line 16:
```suggestion
# Replace [version_manager] with your node version manager (e.g. nvm, nvs, etc.)
```

IchHabeHunger54 (changes requested):
One small typo, looks good otherwise.

README.md
```diff
@@ -2,14 +2,22 @@
...
-The documentation is built using [Docusaurus 2](https://docusaurus.io)
+The documentation is built using [Docusaurus 3](https://docusaurus.io)
 
 ## Contributing
 
 You can read the [contribution guidelines on the docs](https://docs.neoforged.net/contributing/).
 
 If you wish to contribute to the documentation, fork and clone this repository.
 
+The documentation uses Node.js 18. This can either be install manually or using a version manager that supports `.node-version` or `.nvmrc`. One or both of the following commands can be used to install and use the correct Node.js version:
```
XFactHD, line 13:
```suggestion
The documentation uses Node.js 18. This can either be installed manually or using a version manager that supports `.node-version` or `.nvmrc`. One or both of the following commands can be used to install and use the correct Node.js version:
```

README.md
ChampionAsh5357: This is the correct spelling.

README.md
ChampionAsh5357: Whoops, grammar

README.md
```diff
@@ -2,14 +2,22 @@
...
 You can read the [contribution guidelines on the docs](https://docs.neoforged.net/contributing/).
 
 If you wish to contribute to the documentation, fork and clone this repository.
 
+The documentation uses Node.js 18. This can either be installed manually or using a version manager that supports `.node-version` or `.nvmrc`. One or both of the following commands can be used to install and use the correct Node.js version:
+
+```bash
+# Replace [version_manager] with your node version manager (e.g., nvm, nvs, etc.)
+[version_manager] install
+[version_manager] use
```
sciwhiz12, line 18: Just to be clear, we've tested that `nvm` and `nvs` use these commands as stated?

Perhaps it would be better to explicitly mention both commands in-line, and have the text explanation suffice for those using other version managers. (As in, `nvm use   # or nvs use`)

ChampionAsh5357: Based on my review across the different version managers, some of them either require you to only run `install`, `use`, or both. For example, `nvm install` is all that's needed since it will automatically set the newly installed version as the one to use. Though, it doesn't hurt to run `nvm use` if you are switching between multiple versions. `nvs`, on the other hand, uses `use` to automatically install and switch versions. However, you can run `nvs install` and then `nvs use` as well.

That was the main reason I was being so general, but I can give specific examples instead if that's what is wanted.

sciwhiz12 (approved):

XFactHD (approved):

IchHabeHunger54 (approved):

PR #80, opened 2024-05-21
Buffer, no wait, Network...ah, Stream Codecs!

ChampionAsh5357:
Adds documentation on stream codecs and provides links from other references. Tries to keep the page similarly in structure to the codec page for ease of understanding.

------------------
Preview URL: https://pr-80.neoforged-docs-previews.pages.dev


docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,368 @@
...
+    );
+```
+### Lazy Initialized
+
+Sometimes, a stream codec may rely on data that is not present when it is constructed. In these situations `NeoForgeStreamCodecs#lazy` can be used to for a stream codec to construct itself on first read/write. The method takes in a supplied stream codec.
+
+```java
+public static final StreamCodec<ByteBuf, IEventBus> LAZY_STREAM_CODEC = 
+    NeoForgeStreamCodecs.lazy(
+        () -> StreamCodec.Unit(NeoForge.EVENT_BUS)
```
Spinoscythe, line 188:
```suggestion
        () -> StreamCodec.unit(NeoForge.EVENT_BUS)
```

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+The `mapStream` method transforms the buffer using a function that takes in the new buffer type and returns the current buffer type. This method should rarely be used as most methods with stream codecs do not need to change the type of the buffer.
+
+```java
+public static final StreamCodec<RegistryFriendlyByteBuf, Integer> STREAM_CODEC =
+    ByteBufCodecs.VAR_INT.mapStream(buffer -> (ByteBuf) buffer);
+```
+
+### Unit
+
+A stream codec which supplies an in-code value and encodes to nothing can be represented using `StreamCodec#unit`. This is useful if no information should be synced across the network.
```
XFactHD, line 173: It would probably be good to mention that the unit encoder does an equality check on the incoming value and throws if it's not equal to the value given when the unit codec is constructed. This means that non-record types without a valid `equals()` implementation require always passing the *exact* instance instead of a technically equivalent but newly constructed object.

Also, I don't like the example, an event bus seems like a *really* weird thing to put in a packet.

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+
+```java
+public static final StreamCodec<ByteBuf, IEventBus> UNIT_STREAM_CODEC =
+    StreamCodec.unit(
+        () -> NeoForge.EVENT_BUS // Can also be a raw value
```
XFactHD, line 178: There is no supplier overload for the unit `StreamCodec`
```suggestion
        NeoForge.EVENT_BUS
```

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+    );
+```
+### Lazy Initialized
+
+Sometimes, a stream codec may rely on data that is not present when it is constructed. In these situations `NeoForgeStreamCodecs#lazy` can be used to for a stream codec to construct itself on first read/write. The method takes in a supplied stream codec.
```
XFactHD, line 183:
```suggestion
Sometimes, a stream codec may rely on data that is not present when it is constructed. In these situations `NeoForgeStreamCodecs#lazy` can be used for a stream codec to construct itself on first read/write. The method takes in a supplied stream codec.
```

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+// For some enum
+public enum ExampleIdObject {
+    ;
+
+    // Gets Id -> Enum
+    public static final IntFunction<ExampleIdObject> BY_ID = 
+        ByIdMap.continuous(
+            ExampleIdObject::getId,
+            ExampleIdObject.values(),
+        ByIdMap.OutOfBoundsStrategy.ZERO
```
XFactHD, line 266:
```suggestion
            ByIdMap.OutOfBoundsStrategy.ZERO
```

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+A stream codec for sending an `Optional` wrapped value can be generated by supplying a stream codec to `ByteBufCodecs#optional`. This method first reads/writes a boolean indicating whether to read/write the object.
+
+```java
+public static final StreamCodec<RegistryFriendlyByteBuf, Optional<DataComponentType<?>>> OPTIONAL_STREAM_CODEC =
+    DataComponentType.STREAM_CODEC.apply(ByteBufCodecs::optional);
+```
+
+### Registry Objects
+
+Registry objects can be sent across the network using one of three methods: `registry`, `holderRegistry`, or `holder`. Each takes in a `ResourceKey` representing the registry the registry object is in.
```
XFactHD, line 292: This should mention that custom registries must be marked as synced for the first two to work at all and the last one to work with non-direct holders, otherwise the encoder throws an exception (added by Neo).

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+```java
+public static final StreamCodec<RegistryFriendlyByteBuf, Holder<SoundEvent>> STREAM_CODEC =
+    ByteBufCodecs.holder(
+        Registries.SOUND_EVENT, SoundEvent.DIRECT_STREAM_CODEC
+    );
+```
+
+### Holder Sets
+
+Tags or sets of holder wrapped registry objects can be sent using `holderSet`. This takes in a `ResourceKey` represent the registry the registry objects are in.
```
XFactHD, line 317:
```suggestion
Tags or sets of holder wrapped registry objects can be sent using `holderSet`. This takes in a `ResourceKey` representing the registry the registry objects are in.
```

This should mention that for actual tags the tag name is sent and for non-tag holder sets a `StreamCodec` constructed through `holderRegistry()` is used. This also means that the same rules regarding synced registries apply.

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+
+```java
+public static final StreamCodec<RegistryFriendlyByteBuf, HolderSet<Item>> HOLDER_SET_STREAM_CODEC =
+    BytebufCodecs.holderSet(Registries.ITEM);
+```
+
+### Recursive
+
+Sometimes, an object may reference an object of the same type as a field. For example, `MobEffectinstance` takes in an optional `MobEffectinstance` if there is a hidden effect. In this case, `StreamCodec#recursive` can be used to supply the stream codec as part of a function to create the stream codec.
```
XFactHD, line 326:
```suggestion
Sometimes, an object may reference an object of the same type as a field. For example, `MobEffectInstance` takes in an optional `MobEffectInstance` if there is a hidden effect. In this case, `StreamCodec#recursive` can be used to supply the stream codec as part of a function to create the stream codec.
```

XFactHD (changes requested):

docs/networking/streamcodecs.md
```diff
@@ -0,0 +1,369 @@
...
+
+Additionally, there are some static instances that encode and decode primivites and objects using a different method.
+
+#### Unsigned Shorts
+
+`UNSIGNED_SHORT` is an alternative of `SHORT` that is meant to be treated as an unsigned number. As numbers are signed in Java, unsigned shorts are sent and received as `Integer`s with the upper two bytes masked out.
+
+#### Variable-Sized Number
+
+`VAR_INT` and `VAR_LONG` are alternatives of `INT` where the value is encoded to be as small as possible. This is done by encoding seven bits at a time, using the upper bit as a marker of whether there is more data for this number. Numbers between 0 and 2^28-1 for integers or 0 and 2^56-1 for longs will be sent shorter or equal to the number of bytes in a integer or long, respectively. If the values of your numbers are normally between this range, these variable stream codecs should be used.
```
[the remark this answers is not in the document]
ChampionAsh5357: Don't think `LONG` is a thing, but I can rewrite it better.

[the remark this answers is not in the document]
XFactHD: Whoops, should have checked properly, I was pretty sure I remembered that existing 😅

docs/networking/streamcodecs.md
```diff
@@ -60,11 +60,15 @@ Additionally, there are some static instances that encode and decode primivites
...
+
+:::note
+`VAR_INT` is an alternative for `INT`.
+:::
 
 #### Trusted Tags
 
-`TRUSTED_TAG` and `TRUSTED_COMPOUND_TAG` are variants of `TAG` and `COMPOUND_TAG`, respectively, that have an unlimited heap to decode the tag to, compared to the 2MiB limit of `TAG` and `COMPOUND_TAG`. Only the [block entity data packet][blockentity] and [entity data serializers][entityserializer] use these trusted variants.
+`TRUSTED_TAG` and `TRUSTED_COMPOUND_TAG` are variants of `TAG` and `COMPOUND_TAG`, respectively, that have an unlimited heap to decode the tag to, compared to the 2MiB limit of `TAG` and `COMPOUND_TAG`. Trusted tag stream condecs should ideally only be used in clientbound packets, such as what Vanilla does for [block entity data packet][blockentity] and [entity data serializers][entityserializer].
```
XFactHD, line 71:
```suggestion
`TRUSTED_TAG` and `TRUSTED_COMPOUND_TAG` are variants of `TAG` and `COMPOUND_TAG`, respectively, that have an unlimited heap to decode the tag to, compared to the 2MiB limit of `TAG` and `COMPOUND_TAG`. Trusted tag stream codecs should ideally only be used in clientbound packets, such as what Vanilla does for [block entity data packet][blockentity] and [entity data serializers][entityserializer].
```

docs/networking/streamcodecs.md
```diff
@@ -172,20 +176,22 @@ public static final StreamCodec<RegistryFriendlyByteBuf, Integer> STREAM_CODEC =
 
 A stream codec which supplies an in-code value and encodes to nothing can be represented using `StreamCodec#unit`. This is useful if no information should be synced across the network.
 
+:::warning
+Unit stream codecs expects that any encoded object must match the unit specified; otherwise an error will be thrown. Therefore, all objects must have some `equals` implementation that returns true for the unit object.
```
XFactHD, line 180:
```suggestion
Unit stream codecs expect that any encoded object must match the unit specified; otherwise an error will be thrown. Therefore, all objects must have some `equals` implementation that returns true for the unit object.
```
That's not quite true. Using objects with default equality behaviour (i.e. reference equality) is fine, just need to make sure to use the specific instance in every case.

XFactHD (changes requested):

docs/networking/streamcodecs.md
ChampionAsh5357: Yes, though my general thought process is that if you're supplying something within a stream codec, it is unlikely going to be the same specific instance; though I could just mention it still.

XFactHD (approved):

XFactHD (approved):

PR #96, opened 2024-06-02
Update Block Breaking Stage Pipeline

ChampionAsh5357:
Updates the breaking stage pipeline to match what is currently within 1.20.6. Also add a bit more information about the events fired.

------------------
Preview URL: https://pr-96.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/blocks/index.md
+++ b/docs/blocks/index.md
@@ -186,7 +186,12 @@ The following subsections further break down these stages into actual method cal
 
 #### The "Actually Breaking" Stage
 
-- `IItemExtension#onBlockStartBreak` is called. If it returns `true` (determining that the block should not be broken), the pipeline moves to the "finishing" stage.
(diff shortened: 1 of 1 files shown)


Shadows-of-Fire:
There will be incoming changes to this area when I rebase the BlockDropsEvent PR against 1.20.6; not sure if you want to wait for that or adjust it after the fact

ChampionAsh5357:
> There will be incoming changes to this area when I rebase the BlockDropsEvent PR against 1.20.6; not sure if you want to wait for that or adjust it after the fact

I don't think that will change this pipeline in terms of how things are executed. It will only add an additional event line. So, unless the event is coming out within the next few days, I think just judge it on the current structure of NeoForge.

IchHabeHunger54 (approved):

ChampionAsh5357:
Bundled in #108

PR #103, opened 2024-06-07
Update to new enum extension system

XFactHD:
This PR updates the documentation of enum extension to the new system to be implemented in neoforged/FancyModLoader#148

------------------
Preview URL: https://pr-103.neoforged-docs-previews.pages.dev


docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-        throw new IllegalStateException("Enum not extended");
-    }
-
-    // Matches the second constructor
-    public static ExampleEnum create(String name, int arg1, String arg2) {
-        throw new IllegalStateException("Enum not extended");
-    }
-
-    // ...
+To create new enum entries, a JSON file needs to be created and referenced in the `neoforge.mods.toml` with the `enumExtender` entry of a `[[mods]]` block. The specified path must be relative to the `resources` directory.
```
IchHabeHunger54, line 16: Example with a recommended file location here please.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+
+The definition of the entry consists of the target enum's class name, the new field's name (must be prefixed with the mod ID), the descriptor of the constructor to use for constructing the entry and the parameters to be passed to said constructor.
+
+```json5
```
IchHabeHunger54, line 20: Reformat JSON to 2 spaces.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+        });
     }
 }
 ```
 
-### Codecs
+#### Constructor
 
-[Enum codecs][codec] generally take in the `values` array and resolve the codec instantly. As the array is evaluated before any mod entries are added, no mod entries will be supported. There are two solutions to this problem, which to choose depends on whether the enum implements the `StringRepresentable` interface or not.
+The constructor must be specified as a method descriptor and must only contain the parameters visible in the source code, omitting the hidden constant name and ordinal parameters
```
IchHabeHunger54, line 77:
```suggestion
The constructor must be specified as a method descriptor and must only contain the parameters visible in the source code, omitting the hidden constant name and ordinal parameters.
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+                "field": "getTestDamageEffectsParameter"
+            }
+        },
+        {
+            "enum": "net/minecraft/world/item/ItemDisplayContext",
+            "name": "EXAMPLEMOD_STANDING",
+            "constructor": "(ILjava/lang/String;Ljava/lang/String;)V",
+            // Constant parameters provided directly.
+            // This enum has an ID parameter which enforces providing -1 in the case of constant parameters
+            // This enum has a name which enforces prefixing with the mod ID
```
IchHabeHunger54, line 52: I would argue this to be unnecessary context here.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-```java
-// For some enum with codec
-public enum ExampleEnumCodec {
-    // ...
-    ;
+The parameters can be specified in three ways with limitations depending on the parameter types:
 
-    public static final Codec<ExampleEnumCodec> CODEC = Codec.of(/* ... */);
-    public static final StreamCodec<ByteBuf, ExampleEnumCodec> CODEC = StreamCodec.of(/* ... */);
+- Inline in the JSON file as an array of constants (only allowed for primitive values, Strings and for passing null to any reference type)
```
IchHabeHunger54, line 83:
```suggestion
- Inline the parameter values in the JSON file as an array of constants. Only allowed for primitive values, strings and for passing null to any reference type.
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+- As a reference to a field of type `EnumProxy<TheEnum>` in a class from the mod
```
IchHabeHunger54, line 84:
```suggestion
- Specify a reference to a field of type `EnumProxy<TheEnum>` in a class from the mod (see `EnumProxy` example above).
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+- As a reference to a method taking a parameter index and a `Class` object as parameters and returning `Object`
+  - The `Class` object is the type expected for the given parameter and should be used to cast (`Class#cast()`) the return value in order to keep `ClassCastException`s in mod code
```
IchHabeHunger54, line 86:
```suggestion
- Specify a reference to a method returning `Object`, where the return value is the parameter value to use. The method must have exactly two parameters of type `int` (index of the parameter) and `Class<?>` (expected type of the parameter) that should be used to determine the parameter value.
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-    // ...
-}
-
-// The final extended enum may look something like this
-public enum ExampleEnumCodec implements net.neoforged.neoforge.common.IExtensibleEnum {
-    // ...
-    ;
+:::warning
+The fields and/or methods used as parameter sources should be in a separate class to avoid unintentionally loading mod classes too early
+:::
```
IchHabeHunger54, line 90:
```suggestion
:::tip
The fields and/or methods used as sources for parameter values should be in a separate class to avoid unintentionally loading mod classes too early.
:::
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-    public static final StreamCodec<ByteBuf, ExampleEnumCodec> CODEC = 
-        net.neoforged.neoforge.common.util.NeoForgeStreamCodecs.lazy(
-            () -> StreamCodec.of(/* ... */)
-        );
+Certain parameters have additional rules:
 
-    // ...
-}
-```
+- If the parameter is an int ID parameter related to a `@IndexedEnum` annotation on the enum, then it is ignored and replaced by the entry's ordinal. If said parameter is specified inline in the JSON, then it must be specified as `-1`
```
IchHabeHunger54, line 94:
```suggestion
- If the parameter is an int ID parameter related to a `@IndexedEnum` annotation on the enum, then it is ignored and replaced by the entry's ordinal. If said parameter is specified inline in the JSON, then it must be specified as `-1`.
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+- If the parameter is a String name parameter related to a `@NamedEnum` annotation on the enum, then it must be prefixed by the mod ID in the `namespace:path` format known from `ResourceLocation`s
```
IchHabeHunger54, line 95:
```suggestion
- If the parameter is a String name parameter related to a `@NamedEnum` annotation on the enum, then it must be prefixed by the mod ID in the `namespace:path` format known from `ResourceLocation`s.
```
Maybe also add a link to the resource locations article.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
 
-For `StringRepresentable` enums, `IExtensibleEnum` provides `createCodecForExtensibleEnum` and `createStreamCodecForExtensibleEnum` which do not cache the values at all, meaning they will always obtain up-to-date enums added at any point.
+#### Retrieving the generated constant
```
IchHabeHunger54, line 97:
```suggestion
#### Retrieving the Generated Constant
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
 
-```java
-// For some enum with codec
-public enum ExampleEnumStringCodec implements StringRepresentable {
-    // ...
-    ;
+The generated enum constant can be retrieved via `TheEnum.valueOf(String)`. If a field reference is used to provide the parameters, then the constant can also be retrieved from the `EnumProxy` object via `EnumProxy#getValue()`
```
IchHabeHunger54, line 99:
```suggestion
The generated enum constant can be retrieved via `TheEnum.valueOf(String)`. If a field reference is used to provide the parameters, then the constant can also be retrieved from the `EnumProxy` object via `EnumProxy#getValue()`.
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
 
-    ExampleEnumStringCodec(int id, String serializedName) {
-        // ...
-    }
+To add a new extensible enum to NeoForge, there are at least two required things to do:
 
-    // ...
-}
+- First, have the enum implement `IExtensibleEnum` to mark that this enum should be transformed via the `RuntimeEnumExtender`
+- Second, add a `getExtensionInfo` method which returns `ExtensionInfo.nonExtended(TheEnum.class)`
```
IchHabeHunger54, line 106:
```suggestion
- Make the enum implement `IExtensibleEnum` to mark that this enum should be transformed via the `RuntimeEnumExtender`.
- Add a `getExtensionInfo` method that returns `ExtensionInfo.nonExtended(TheEnum.class)`.
```

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,134 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-    ;
+Further action is required depending on specific details about the enum:
 
-    private static final java.util.Map<String, ExampleEnumStringCodec> BY_NAME = 
-        java.util.Arrays.stream(ExampleEnumStringCodec.values())
-        .collect(java.util.stream.Collectors.toMap(
-            e -> e.serializedName, e -> e
-        ));
+- If the enum has an int ID parameter which should match the entry's ordinal, then the enum should be annotated with `@NumberedEnum` with the ID's parameter index as the annotation's value if it's not the first parameter
+- If the enum has a String name parameter which is used for serialization and should therefore be namespaced, then the enum should be annotated with `@NamedEnum` with the name's parameter index as the annotation's value if it's not the first parameter
```
IchHabeHunger54, line 111: I am confused as to what the annotation parameters are supposed to be, needs improved wording.

IchHabeHunger54 (changes requested):
Mostly formatting or rewording changes, looking good otherwise.

docs/advanced/extensibleenums.md
XFactHD: I disagree with making this a tip, it's not merely a nice-to-have.

docs/advanced/extensibleenums.md
XFactHD: The annotation parameter is the relevant constructor parameter's index. In most cases that would be 0 (the annotation's default value), except when the given constructor parameter is not the first one (i.e. when an enum has an ID and a name parameter as the first and second parameter respectively, then index 1 needs to be specified in the annotation for the latter). Not quite sure at the moment how to word that better.

docs/advanced/extensibleenums.md
XFactHD: I'm not a fan of that wording in a listing. I've reworded the points a bit, let me know if that's better.

IchHabeHunger54 (dismissed):
looks good enough to me, shipit

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,139 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+- If the enum is sent over the network, then it should be annotated with `@NetworkedEnum` with the annotation's parameter specifying in which direction the values may be sent (clientbound, serverbound or bidirectional)
+  - TODO: add to network check if necessary
```
ChampionAsh5357, line 118: Ah yes, a lovely todo

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,139 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-// In your main mod class
-public static final FireworkExplosion.Shape DUMMY =
-        FireworkExplosion.Shape.create("EXAMPLEMOD_DUMMY", 294, "examplemod_dummy");
+To create new enum entries, a JSON file needs to be created and referenced in the `neoforge.mods.toml` with the `enumExtensions` entry of a `[[mods]]` block. The specified path must be relative to the `resources` directory:
+```toml
+# In neoforge.mods.toml:
+[[mods]]
+## The file is relative to the output directory of the resources, or the root path inside the jar when compiled
+## The 'resources' directory represents the root output directory of the resources
+enumExtensions="META-INF/enumextensions.json"
```
ChampionAsh5357, line 22: I think this also needs to be added to modfiles.md as a new option as well. That can probably link to this page.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,139 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+        });
     }
 }
 ```
 
-### Codecs
+#### Constructor
 
-[Enum codecs][codec] generally take in the `values` array and resolve the codec instantly. As the array is evaluated before any mod entries are added, no mod entries will be supported. There are two solutions to this problem, which to choose depends on whether the enum implements the `StringRepresentable` interface or not.
+The constructor must be specified as a method descriptor and must only contain the parameters visible in the source code, omitting the hidden constant name and ordinal parameters.
```
ChampionAsh5357, line 82: Probably a link to the JVMS page similar to Access transformers on how to write out the definition.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,139 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-    public static final Codec<ExampleEnumCodec> CODEC = Codec.of(/* ... */);
-    public static final StreamCodec<ByteBuf, ExampleEnumCodec> CODEC = StreamCodec.of(/* ... */);
+The parameters can be specified in three ways with limitations depending on the parameter types:
 
-    // ...
-}
+- Inline in the JSON file as an array of constants (only allowed for primitive values, Strings and for passing null to any reference type)
+- As a reference to a field of type `EnumProxy<TheEnum>` in a class from the mod (see `EnumProxy` example above)
+- As a reference to a method returning `Object`, where the return value is the parameter value to use. The method must have exactly two parameters of type `int` (index of the parameter) and `Class<?>` (expected type of the parameter)
+  - The `Class<?>` object should be used to cast (`Class#cast()`) the return value in order to keep `ClassCastException`s in mod code.
```
ChampionAsh5357, line 91: I would order these how you order them in the above example. Or, it would probably be better to construct a separate example representing how each is used, because this feels a bit too brief.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,139 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+:::
 
-    // ...
-}
-```
+Certain parameters have additional rules:
 
-For `StringRepresentable` enums, `IExtensibleEnum` provides `createCodecForExtensibleEnum` and `createStreamCodecForExtensibleEnum` which do not cache the values at all, meaning they will always obtain up-to-date enums added at any point.
+- If the parameter is an int ID parameter related to a `@IndexedEnum` annotation on the enum, then it is ignored and replaced by the entry's ordinal. If said parameter is specified inline in the JSON, then it must be specified as `-1`.
+- If the parameter is a String name parameter related to a `@NamedEnum` annotation on the enum, then it must be prefixed by the mod ID in the `namespace:path` format known from `ResourceLocation`s.
```
ChampionAsh5357, line 100: Does this throw an error if not prefixed?

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,139 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+The fields and/or methods used as sources for parameter values should be in a separate class to avoid unintentionally loading mod classes too early.
+:::
 
-    // ...
-}
-```
+Certain parameters have additional rules:
 
-For `StringRepresentable` enums, `IExtensibleEnum` provides `createCodecForExtensibleEnum` and `createStreamCodecForExtensibleEnum` which do not cache the values at all, meaning they will always obtain up-to-date enums added at any point.
+- If the parameter is an int ID parameter related to a `@IndexedEnum` annotation on the enum, then it is ignored and replaced by the entry's ordinal. If said parameter is specified inline in the JSON, then it must be specified as `-1`.
```
ChampionAsh5357, line 99: Does this throw an error if not -1? And how does this work if the index enum is not the first parameter (I don't think this happens in vanilla, but I'm unsure whether ignore means only if its the first parameter).

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,139 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-        .collect(java.util.stream.Collectors.toMap(
-            e -> e.serializedName, e -> e
-        ));
+- If the enum has an int ID parameter which should match the entry's ordinal, then the enum should be annotated with `@NumberedEnum` with the ID's parameter index as the annotation's value if it's not the first parameter
+- If the enum has a String name parameter which is used for serialization and should therefore be namespaced, then the enum should be annotated with `@NamedEnum` with the name's parameter index as the annotation's value if it's not the first parameter
+- If the enum is sent over the network, then it should be annotated with `@NetworkedEnum` with the annotation's parameter specifying in which direction the values may be sent (clientbound, serverbound or bidirectional)
+  - Warning: networked enums will require additional steps once network checks for enums are implemented in NeoForge
+- If the enum has constructors which are not usable by mods (i.e. because they require registry objects on an enum that may be initialized before modded registration runs), then they should be annotated with `@ReservedConstructor`
```
ChampionAsh5357, line 119: This should also be explained above when creating an extensible enum since this means that the constructor cannot be selected.

ChampionAsh5357 (changes requested):
Great! General comments for more info.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,142 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+      }
+    },
+    {
+      "enum": "net/minecraft/world/damagesource/DamageEffects",
+      "name": "EXAMPLEMOD_TEST",
+      "constructor": "(Ljava/lang/String;Ljava/util/function/Supplier;)V",
+      // The parameters to be used, provided as a reference to a method in the given class
+      "parameters": {
+        "class": "example/examplemod/MyEnumParams",
+        "field": "getTestDamageEffectsParameter"
```
SizableShrimp, line 57: Without looking at the system, I'm going to assume this is incorrect because the key is "field", and this is referencing a method.

XFactHD: Indeed, copy-paste screwed me over once again

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,142 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
+    },
+    {
+      "enum": "net/minecraft/world/damagesource/DamageEffects",
+      "name": "EXAMPLEMOD_TEST",
+      "constructor": "(Ljava/lang/String;Ljava/util/function/Supplier;)V",
+      // The parameters to be used, provided as a reference to a method in the given class
+      "parameters": {
+        "class": "example/examplemod/MyEnumParams",
+        "method": "getTestDamageEffectsParameter"
+      }
```
ChampionAsh5357, line 58: Four spaces per indentation.

docs/advanced/extensibleenums.md
```diff
@@ -7,216 +7,142 @@ Extensible Enums are an enhancement of specific Vanilla enums to allow new entri
...
-    // ...
-    ;
+The parameters can be specified in three ways with limitations depending on the parameter types:
 
-    public static final Codec<ExampleEnumCodec> CODEC = Codec.of(/* ... */);
-    public static final StreamCodec<ByteBuf, ExampleEnumCodec> CODEC = StreamCodec.of(/* ... */);
+- Inline in the JSON file as an array of constants (only allowed for primitive values, Strings and for passing null to any reference type)
+- As a reference to a field of type `EnumProxy<TheEnum>` in a class from the mod (see `EnumProxy` example above)
+- As a reference to a method returning `Object`, where the return value is the parameter value to use. The method must have exactly two parameters of type `int` (index of the parameter) and `Class<?>` (expected type of the parameter)
+  - The `Class<?>` object should be used to cast (`Class#cast()`) the return value in order to keep `ClassCastException`s in mod code.
```
ChampionAsh5357, line 92: So, if `Class#cast` should be used to cast the return value, why isn't that done within the example?

ChampionAsh5357 (changes requested):
Looks good. Just a formatting change and a mismatched description/example.

docs/advanced/extensibleenums.md
XFactHD: The example does use it, the switch is wrapped in the `Class#cast()` call. I can extract it to a separate line if you prefer that.

ChampionAsh5357: Ah, my bad. I simply missed it because it was wrapping around the switch statement.

ChampionAsh5357 (approved):

Matyrobbrt (approved):

PR #135, opened 2024-07-13
Fix Access Transformers "Adding ATs" example

ManTouQAQ:


ChampionAsh5357:
This needs quite a bit of work. First, the original example is still correct if you are using NeoGradle. The example you are overwriting is for ModDevGradle, so it should ideally be in its own separate tab. Additionally, the ModDevGradle example is not completely correct either, as the default location for access transformers is automatically detected. If you are planning on adding support for ModDevGradle, it should be done across all gradle examples.

PR #139, opened 2024-07-30
Fix menu constructor example

ChampionAsh5357:
Fixes the menu constructor example by specifying where additional parameters should be added.

------------------
Preview URL: https://pr-139.neoforged-docs-previews.pages.dev


IchHabeHunger54 (approved):
This article is due for a rework anyway, but I think for now we can include this.

PR #140, opened 2024-07-30
Fix particle description information

ChampionAsh5357:
Fixes the particle description info to say that it must be registered for certain provider methods. Also includes an addendum for how the particle description images are used.

------------------
Preview URL: https://pr-140.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/resources/client/particles.md
+++ b/docs/resources/client/particles.md
@@ -120,7 +120,7 @@ Finally, we must associate our particle type with a texture. Similar to how item
 }
 ```
 
-A particle definition are required when using a particle that takes in a `SpriteSet`, which is done when registering a particle provider via `registerSpriteSet` or `registerSprite`. They must **not** be provided for particle providers registered via `#registerSpecial`.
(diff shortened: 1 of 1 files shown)


docs/resources/client/particles.md
```diff
@@ -120,12 +120,16 @@ Finally, we must associate our particle type with a texture. Similar to how item
 }
 ```
 
-Note that a particle definition file is only necessary when using a sprite set particle. Single sprite particles directly map to the texture file at `assets/<namespace>/textures/particle/<particle_name>.png`, and special particle providers can do whatever you want anyway.
+A particle definition are required when using a particle that takes in a `SpriteSet`, which is done when registering a particle provider via `registerSpriteSet` or `registerSprite`. They must **not** be provided for particle providers registered via `#registerSpecial`.
```
XFactHD, line 123:
```suggestion
A particle definition is required when using a particle that takes in a `SpriteSet`, which is done when registering a particle provider via `registerSpriteSet` or `registerSprite`. They must **not** be provided for particle providers registered via `#registerSpecial`.
```

XFactHD (approved):

PR #156, opened 2024-08-19
Fix Network Changes on Payload Thread Handling

### LINKED ISSUE #149: Network/payload docs are wrong in regard to thread handling
Nowadays, the PayloadRegistrar sends all handlers to the main thread by default. However, the documentation (https://docs.neoforged.net/docs/networking/payload/) not only doesn't mention that it can do that but explicitly says the handlers run on the network thread.

ChampionAsh5357:
Closes #149 

Updates the docs to mention that running packets on the main thread is done by default. In addition, add some more information about running the payload on the network thread.

------------------
Preview URL: https://pr-156.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/networking/payload.md
+++ b/docs/networking/payload.md
@@ -119,8 +119,9 @@ PayloadRegistrar registrar = event.registrar("1");
 registrar.playBidirectional(...); // On the main thread
 registrar.playBidirectional(...); // On the main thread
 
-// All registrar methods create a new instance, so the change
-// needs to be updated by storing the result
+// Configuration methods modify the state of the registrar
+// by creating a new instance, so the change needs to be
+/// updated by storing the result
 registrar = registrar.executesOn(HandlerThread.NETWORK);
 
 registrar.playBidirectional(...); // On the network thread
--- a/docs/networking/payload.md
+++ b/docs/networking/payload.md
@@ -89,7 +89,7 @@ public class ClientPayloadHandler {
 Here a couple of things are of note:
 
 - The handling method here gets the payload, and a contextual object.
-- The handler of the payload method is, by default, invoked on the main thread.
+- The handling method of the payload is, by default, invoked on the main thread.
 
 
 If you need to do some computation that is resource intensive, then the work should be done on the network thread, instead of blocking the main thread. This is done by setting the `HandlerThread` of the `PayloadRegistrar` to `HandlerThread#NETWORK` via `PayloadRegistrar#executesOn` before registering the payload.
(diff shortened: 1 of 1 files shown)


docs/networking/payload.md
```diff
@@ -79,7 +79,48 @@ Now that we have registered the payload we need to implement a handler. For this
...
+}
+```
+
+Here a couple of things are of note:
+
+- The handling method here gets the payload, and a contextual object.
+- The handler of the payload method is, by default, invoked on the main thread.
+
+
+If you need to do some computation that is resource intensive, then the work should be done on the network thread, instead of blocking the main thread. This is done by setting the `HandlerThread` of the `PayloadRegistrar` to `HandlerThread#NETWORK` via `PayloadRegistrar#executesOn` before registering the payload.
```
XFactHD, line 95: It may make sense to mention that this is retained for all further registrations until `executesOn()` is called again with a different value.

docs/networking/payload.md
```diff
@@ -79,7 +79,71 @@ Now that we have registered the payload we need to implement a handler. For this
...
+        // Do something with the data, on the main thread
+        blah(data.age());
+    }
+}
+```
+
+Here a couple of things are of note:
+
+- The handling method here gets the payload, and a contextual object.
+- The handler of the payload method is, by default, invoked on the main thread.
```
XFactHD, line 92:
```suggestion
- The handling method of the payload is, by default, invoked on the main thread.
```

docs/networking/payload.md
```diff
@@ -79,7 +79,71 @@ Now that we have registered the payload we need to implement a handler. For this
...
+:::note
+All payloads registered after an `executesOn` call will retain the same thread execution location until `executesOn` is called again.
+
+```java
+PayloadRegistrar registrar = event.registrar("1");
+
+registrar.playBidirectional(...); // On the main thread
+registrar.playBidirectional(...); // On the main thread
+
+// All registrar methods create a new instance, so the change
```
XFactHD, line 122:
```suggestion
// All registrar configuration methods create a new instance, so the change
```
"All methods" sounds like actual registration methods also create a new registrar (which is of course not the case).
Though I'm not quite happy with calling them "configuration methods" either, maybe you can think of something better.

XFactHD (changes requested):

docs/networking/payload.md
ChampionAsh5357: I couldn't think of a better descriptor, so I opted to provide more context instead by referencing the state. Let me know if this is better.

IchHabeHunger54 (approved):

PR #160, opened 2024-09-11
Rewrite the Block Entities section

IchHabeHunger54:
Rewrites everything block-entity related to cover the following topics in more detail and with up-to-date info. Specifically, it:

- Rewrites the Block Entity article to be more beginner-friendly and use more examples
- Mentions that data attachments are available on block entities
- Adds an article about containers
- Rewords the BER article
- Merges the BEWLR article into the BER article, and rewords it

------------------
Preview URL: https://pr-160.neoforged-docs-previews.pages.dev


docs/blockentities/container.md
```diff
@@ -0,0 +1,250 @@
...
+    // We again defer to ContainerHelper here.
+    @Override
+    public ItemStack removeItemNoUpdate(int slot) {
+        return ContainerHelper.takeItem(items, slot);
+    }
+
+    // Set the given item stack in the given slot.
+    @Override
+    public void setItem(int slot, ItemStack stack) {
+        items.set(slot, stack);
```
ChampionAsh5357, line 59: Should respect the max stack size that can be accepted by the container. `stack.limitSize(this.getMaxStackSize(stack));`

docs/blockentities/container.md
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    // Do something when changes are done to the container. For example, you could call BlockEntity#setChanged here.
+    @Override
+    public void setChanged() {
```
ChampionAsh5357, line 64: This should be called whenever mutating the inventory (e.g., setItem, removeItem) since you're giving an example.

docs/blockentities/container.md
```diff
@@ -0,0 +1,250 @@
...
+
+// Set the first item in the container to dirt.
+container.setItem(0, new ItemStack(Items.DIRT));
+
+// Removes a quantity of (up to) 16 from the third slot.
+container.removeItem(2, 16);
+```
+
+:::warning
+A container may throw an exception if trying to access a slot that is beyond its container size.
```
ChampionAsh5357, line 170: A bit misleading. For most implementations, an exception will be thrown. This only doesn't happen for `SimpleContainer` as it will return an empty stack.

docs/blockentities/container.md
```diff
@@ -0,0 +1,250 @@
...
+    public MyBackpackContainer(ItemStack stack) {
+        // We call super with our desired container size.
+        super(27);
+        // Setting the stack field.
+        this.stack = stack;
+        // We load the container contents from the data component (if present), which is represented
+        // by the ItemContainerContents class. If absent, we use ItemContainerContents.EMPTY.
+        ItemContainerContents contents = stack.getOrDefault(DataComponents.CONTAINER, ItemContainerContents.EMPTY);
+        // Copy the data component contents into our item stack list.
+        contents.copyInfo(this.getItems());
```
ChampionAsh5357, line 193:
```suggestion
        contents.copyInto(this.getItems());
```

docs/blockentities/container.md
```diff
@@ -0,0 +1,250 @@
...
+:::
+
+## `Container`s on `ItemStack`s
+
+Until now, we mainly discussed `Container`s on `BlockEntity`s. However, they can also be applied to [`ItemStack`s][itemstack] using the `minecraft:container` [data component][datacomponent]:
+
+```java
+// We use SimpleContainer as the superclass here so we don't have to reimplement the item handling logic ourselves.
+// You may of course use a different base implementation of Container (or implement Container yourself) if needed.
+public class MyBackpackContainer extends SimpleContainer {
```
ChampionAsh5357, line 180: This is very wonky as because the component maintains a copy of the stacks on update, this will not reflect any changes from someone else operating on the component at the same time. My immediate thought is the bundle, as that would be interacting with the data while in a container, which could be accessed by multiple people.

Not really sure what would be the best way to handle this, unless you are also getting the items from the container contents based on the provided slot.

docs/blockentities/container.md
```diff
@@ -0,0 +1,250 @@
...
+}
+```
+
+Keep in mind that this class is a `BlockEntity` and a `Container` at the same time. This means that you can use the class as a supertype for your block entity to get a functioning block entity with a pre-implemented container.
+
+## Using Containers
+
+Now that we have created containers, let's use them!
+
+Since there is a considerable overlap between `Container`s and `BlockEntity`s, containers are best retrieved by casting the block entity to `Container` if possible:
```
ChampionAsh5357, line 148: Should probably talk about `WorldlyContainer` here as well, since that allows access through the container based upon the interacted side.

ChampionAsh5357 (changes requested):

docs/blockentities/ber.md
```diff
@@ -1,44 +1,102 @@
...
+A `BlockEntityRenderer`, often abbreviated as BER, is used to render [blocks][block] in a way that cannot be represented with a [static baked model][model] (JSON, OBJ, others). For example, this could be used to dynamically render container contents of a chest-like block. A block entity renderer requires the block to have a [`BlockEntity`][blockentity], even if the block does not store any data otherwise.
 
-## Creating a BER
+To create a BER, create a class that inherits from `BlockEntityRenderer`. It takes a generic argument specifying the block's `BlockEntity` class, which is used as a parameter type in the BER's `render` method.
 
-To create a BER, create a class that inherits from `BlockEntityRenderer`. It takes a generic argument specifying the block's `BlockEntity` class. The generic argument is used in the BER's `render` method.
+```java
+// Assumes the existence of MyBlockEntity as a subclass of BlockEntity.
+public class MyBlockEntityRenderer implements BlockEntityRenderer<MyBlockEntity> {
```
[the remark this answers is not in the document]
IchHabeHunger54: Done, I still kept an example of omitting the constructor below.

docs/blockentities/ber.md
```diff
@@ -1,44 +1,102 @@
...
     }
+}
+```
+
+Keep in mind that, like with BERs, there is only one instance of your BEWLR. Stack-specific properties should therefore be stored in the stack, not the BEWLR.
+
+Unlike BERs, we do not register BEWLRs directly. Instead, we register an instance of `IClientItemExtensions` to the `RegisterClientExtensionsEvent`. `IClientItemExtensions` allows us to specify a number of rendering-related behaviors on items, but since we're only interested in replacing the renderer with our newly-made BEWLR, we will just slap the `IClientItemExtensions` interface onto our BEWLR, like so:
+
+```java
+public class MyBlockEntityWithoutLevelRenderer extends BlockEntityWithoutLevelRenderer implements IClientItemExtensions {
```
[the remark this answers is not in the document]
IchHabeHunger54: I use this in my mods without problems. This is still a unique instance after all; we only ever create one instance of this class during the RegisterClientExtensionsEvent.

docs/blockentities/container.md
```diff
@@ -0,0 +1,250 @@
+# Containers
```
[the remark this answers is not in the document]
IchHabeHunger54: Generally yes, I'd just not add a note until we actually have an article on those (which I'll postpone to a separate PR).

docs/blockentities/container.md
IchHabeHunger54: I added a note that race conditions can happen this way, and just assumed that only one player can access the backpack on this example.

docs/blockentities/container.md
IchHabeHunger54: Added a WorldlyContainer section.

docs/blockentities/index.md
```diff
@@ -1,141 +1,237 @@
...
+
+public static final Supplier<BlockEntityType<MyBlockEntity>> MY_BLOCK_ENTITY = BLOCK_ENTITY_TYPES.register(
+        "my_block_entity",
+        // The block entity type, created using a builder.
+        () -> BlockEntityType.Builder.of(
+                // The supplier to use for constructing the block entity instances.
+                MyBlockEntity::new,
+                // A vararg of blocks that can have this block entity.
+                // This assumes the existence of the referenced blocks as DeferredBlock<Block>s.
+                MyBlocks.MY_BLOCK_1, MyBlocks.MY_BLOCK_2
```
ChampionAsh5357, line 39: One of these blocks should match the example block you just created below.

docs/blockentities/container.md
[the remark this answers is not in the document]
ChampionAsh5357: I would write this up as an issue so that we don't forget it.

docs/blockentities/container.md
ChampionAsh5357: I was referring to calling this literally in `setItem` and so forth similar to how `SimpleContainer` does it as this method is supposed to listen to when the inventory is changed.

docs/blockentities/container.md
ChampionAsh5357: That seems fine to me. If a better solution comes up later, we can always replace it.

docs/blockentities/ber.md
```diff
@@ -1,44 +1,122 @@
...
+}
+```
+
+Keep in mind that, like with BERs, there is only one instance of your BEWLR. Stack-specific properties should therefore be stored in the stack, not the BEWLR.
 
-    // Implement #render method here and any other logic
+Unlike BERs, we do not register BEWLRs directly. Instead, we register an instance of `IClientItemExtensions` to the `RegisterClientExtensionsEvent`. `IClientItemExtensions` allows us to specify a number of rendering-related behaviors on items, but since we're only interested in replacing the renderer with our newly-made BEWLR, we will just slap the `IClientItemExtensions` interface onto our BEWLR, like so:
+
+```java
+public class MyBlockEntityWithoutLevelRenderer extends BlockEntityWithoutLevelRenderer implements IClientItemExtensions {
```
XFactHD, line 89: While this does work just fine, I don't think we should be suggesting this in the docs. Proper separation of concerns should be preferred in docs aimed mostly at beginners.

docs/blockentities/container.md
```diff
@@ -0,0 +1,302 @@
...
+- The `items` list covers the 36 main inventory slots, including the nine hotbar slots (indices 0-8).
+- The `armor` list is a list of length 4, containing armor for the `FEET`, `LEGS`, `CHEST`, and `HEAD`, in that order. This list uses `EquipmentSlot` accessors, similar to `Mob`s (see above).
+- The `offhand` list contains only the offhand slot, i.e. has a length of 1.
+
+When iterating over the inventory contents, it is recommended to iterate over `items`, then over `armor` and then over `offhand`, to be consistent with vanilla behavior.
+
+[block]: ../blocks/index.md
+[blockentity]: index.md
+[component]: ../resources/client/i18n.md#components
+[datacomponent]: ../items/datacomponents.md
```
XFactHD, line 299: The data components page is an MDX
```suggestion
[datacomponent]: ../items/datacomponents.mdx
```

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
...
+It is important that you do safety checks, as the `BlockEntity` might already be destroyed/replaced when the message arrives at the player. You should also check if the chunk is loaded via `Level#hasChunkAt`.
 :::
 
-[registration]: ../concepts/registries.md#methods-for-registering
-[storing-data]: #storing-data-within-your-blockentity
-[menu]: ../gui/menus.md
+[block]: ../blocks/index.md
+[blockstate]: ../blocks/states.md
+[dataattachments]: ../datastorage/attachments.md
+[datacomponents]: ../items/datacomponents.md
```
XFactHD, line 235: The data components page is an MDX
```suggestion
[datacomponents]: ../items/datacomponents.mdx
```

docs/blockentities/container.md
```diff
@@ -0,0 +1,302 @@
+# Containers
+
+A popular use case of [block entities][blockentity] is to store items of some kind. Some of the most essential [blocks][block] in Minecraft, such as the furnace or the chest, use block entities for this purpose. To store items on something, Minecraft uses `Container`s.
+
+The `Container` interface defines methods such as `#getItem`, `#setItem` and `#removeItem` that can be used to query and update the container. Since it is an interface, it does not actually contain a backing list or other data structure, that is up to the implementing system.
+
+Due to this, `Container`s can not only be implemented on block entities, but any other class as well. Notable examples include entity inventories, as well as common modded [items][item] such as backpacks.
+
+:::info
+NeoForge provides the `ItemStackHandler` class as a replacement for `Container`s in many places. It should be used wherever possible in favor of `Container`, as it allows for cleaner interaction with other `Container`s/`ItemStackHandler`s.
```
XFactHD, line 10: Items themselves should never implement `Container` or anything similar, there is no way to implement the methods in a sane way on an item.

This should also note that other mods should never be expected to implement `Container` or `WorldlyContainer` on any of their inventory-holding things (mainly entities and BEs) and that disregarding the item handler capability severely limits compatibility

docs/blockentities/container.md
```diff
@@ -0,0 +1,302 @@
...
+    // When the contents are changed, we save the data component on the stack.
+    @Override
+    public void setChanged() {
+        super.setChanged();
+        this.stack.set(DataComponents.CONTAINER, ItemContainerContents.fromItems(this.getItems()));
+    }
+}
+```
+
+And voilà, you have created an item-backed container! Simply call `new MyBackpackContainer(stack)` to create a container for a menu or other use case.
```
XFactHD, line 257: This should note that wrapping a data component in a `Container` for use in a menu requires the use of a container-compatible implementation of NeoForge's `StackCopySlot`, otherwise the menu will mutate the data component, breaking the immutability contract of data components.

docs/blockentities/container.md
```diff
@@ -0,0 +1,302 @@
...
+
+## `Container`s on `Entity`s
+
+`Container`s on `Entity`s are finicky: whether an entity has a container or not cannot be universally determined. It all depends on what entity you are handling, and as such can require a lot of special-casing.
+
+If you are creating an entity yourself, there is nothing stopping you from implementing `Container` on it directly, though be aware that you will not be able to use superclasses such as `SimpleContainer` (since `Entity` is the superclass).
+
+### `Container`s on `Mob`s
```
XFactHD, line 265: This should also mention the InventoryCarrier interface which is implemented by some of the vanilla entities (villagers, pillagers, allays, piglins and wandering traders).

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
 # Block Entities
 
-`BlockEntities` are like simplified `Entities` that are bound to a Block. They are used to store dynamic data, execute tick based tasks, and dynamic rendering. Some examples from vanilla Minecraft would be handling of inventories on chests, smelting logic on furnaces, or area effects on beacons. More advanced examples exist in mods, such as quarries, sorting machines, pipes, and displays.
+Block entities allow the storage of data on [blocks][block] in cases where [block states][blockstate] are not suited. This is especially the case for data with a non-finite amount of options, such as inventories. Block entities are stationary and bound to a block, but otherwise share many similarities with entities, hence the name.
```
XFactHD, line 3:
```suggestion
Block entities allow the storage of data on [blocks][block] in cases where [block states][blockstate] are not suitable. This is especially the case for data with a non-finite amount of options, such as inventories. Block entities are stationary and bound to a block, but otherwise share many similarities with entities, hence the name.
```

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
...
+        BLOCKS.register("my_block_1", () -> new MyEntityBlock( /* ... */ ));
+public static final DeferredBlock<MyEntityBlock> MY_BLOCK_2 =
+        BLOCKS.register("my_block_2", () -> new MyEntityBlock( /* ... */ ));
+```
 
-There are three ways of syncing data to the client: synchronizing on chunk load, on block updates, and with a custom network message.
+## Storing Data
 
-### Synchronizing on LevelChunk Load
+One of the main purposes of `BlockEntity`s is to store data. Data storage on block entities can happen in two ways: directly reading and writing [NBT][nbt], or using [data attachments][dataattachments]. This section will cover reading and writing NBT directly; for data attachments, please refer to the linked article.
```
XFactHD, line 89: This should mention that data attachments are almost exclusively relevant to BEs from vanilla or other mods. For a mod's own BEs, direct storage in the BE itself should usually be preferred.

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
...
+In both methods, it is important that you call super, as that adds basic information such as the position. The tag names `id`, `x`, `y`, `z`, `NeoForgeData` and `neoforge:attachments` are reserved by the super methods, and as such, you should not use them yourself.
 
-:::caution
-Synchronizing excessive/useless data for block entities can lead to network congestion. You should optimize your network usage by sending only the information the client needs when the client needs it. For instance, it is more often than not unnecessary to send the inventory of a block entity in the update tag, as this can be synchronized via its [`AbstractContainerMenu`][menu].
+:::info
+It is expected that Mojang will adapt the [Data Components][datacomponents] system to also work with block entities sometime during the next few updates. Once that happens, both saving to NBT and data attachments will be removed in favor of data components.
```
XFactHD, line 123: I would not jump to this conclusion yet. We'll get there when and if it actually happens.

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
...
+// Note: The ticker is defined in the block, not the block entity. However, it is good practice to
+// keep the ticking logic in the block entity in some way, for example by defining a static #tick method.
+public class MyEntityBlock extends Block implements EntityBlock {
+    // other stuff here
+
+    @Override
+    public <T extends BlockEntity> BlockEntityTicker<T> getTicker(Level level, BlockState state, BlockEntityType<T> type) {
+        // You can return different tickers here, depending on whatever factors you want. A common use case would be
+        // to return different tickers on the client or server, or only tick one side to begin with.
+        return type == MY_BLOCK_ENTITY.get() ? MyBlockEntity::tick : null;
```
XFactHD, line 142: Due to the specific generic on the type and ticker, the ticker being returned must be casted to `BlockEntityTicker<T>` with an unchecked cast. This is why vanilla has `BaseEntityBlock.createTickerHelper()` which takes the incoming type, the actual type and the ticker and does this type comparison and cast for you. It's protected however, so one would need to copy or AT it as Neo currently does not provide an AT for it.

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
...
-  return tag;
+// Note: The ticker is defined in the block, not the block entity. However, it is good practice to
+// keep the ticking logic in the block entity in some way, for example by defining a static #tick method.
+public class MyEntityBlock extends Block implements EntityBlock {
+    // other stuff here
+
+    @Override
+    public <T extends BlockEntity> BlockEntityTicker<T> getTicker(Level level, BlockState state, BlockEntityType<T> type) {
+        // You can return different tickers here, depending on whatever factors you want. A common use case would be
+        // to return different tickers on the client or server, or only tick one side to begin with.
```
XFactHD, line 141: This should also mention the ability to only tick for certain `BlockState`s

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
...
+
+```java
+public class MyBlockEntity extends BlockEntity {
+    // ...
+
+    // Create an update tag here. For block entities with only a few fields, this can just call #saveAdditional.
+    @Override
+    public CompoundTag getUpdateTag(HolderLookup.Provider registries) {
+        CompoundTag tag = new CompoundTag();
+        // You can also opt to only save the data that has actually changed here.
```
XFactHD, line 175: `getUpdateTag()` must always send all the data the client needs. Omitting unchanged data is only an option in `getUpdatePacket()` when the packet constructor with the `BiFunction<BlockEntity, RegistryAccess, CompoundTag>` parameter is used.

docs/blockentities/index.md
```diff
@@ -1,141 +1,239 @@
...
+    public CompoundTag getUpdateTag(HolderLookup.Provider registries) {
+        CompoundTag tag = new CompoundTag();
+        saveAdditional(tag, registries);
+        return tag;
+    }
+
+    // Return our packet here. This method returning a non-null result tells the game to use this packet for syncing.
+    @Override
+    public Packet<ClientGamePacketListener> getUpdatePacket() {
+        return ClientboundBlockEntityDataPacket.create(this);
```
XFactHD, line 209: This should mention that the single-arg constructor uses `getUpdateTag()` to collect the data to be synced. The second constructor should be mentioned as well with the additional ability to omit unchanged data as mentioned before.

XFactHD (changes requested):

docs/blockentities/container.md
```diff
@@ -254,7 +256,11 @@ public class MyBackpackContainer extends SimpleContainer {
 }
 ```
 
-And voilà, you have created an item-backed container! Simply call `new MyBackpackContainer(stack)` to create a container for a menu or other use case.
+And voilà, you have created an item-backed container! Call `new MyBackpackContainer(stack)` to create a container for a menu or other use case.
+
+:::warning
+Be aware that `Menu`s that directly interface with `Container`s must `#copy()` their `ItemStack`s when modifying them, as otherwise the immutability contract on data components is broken.
```
XFactHD, line 262: This should specifically refer to `StackCopySlot` as I mentioned here: https://github.com/neoforged/Documentation/pull/160#discussion_r1774248427. There is no reason for mods to reinvent the wheel when NeoForge already provides this.

XFactHD (changes requested):

XFactHD (dismissed):

docs/blockentities/ber.md
```diff
@@ -1,44 +1,119 @@
...
 
-// In another class using some method to listen to this event
+```java
 @SubscribeEvent
-public static void registerRenderers(EntityRenderersEvent.RegisterRenderers event) {
-    event.registerBlockEntityRenderer(MyBlockEntityTypes.MYBE.get(), MyBlockEntityRenderer::new);
+public static void registerClientExtensions(RegisterClientExtensionsEvent event) {
+    event.registerItem(
+            // The only instance of our IClientItemExtensions, and as such, the only instance of our BEWLR.
+            new MyClientItemExtensions(),
```
ChampionAsh5357, line 107: I would add a note that there should only ever be one instance of an extension subclass, as it is essentially treated as a singleton itself where stack methods are routed through.

ChampionAsh5357 (changes requested):
Just one quick comment and then it can be merged.

ChampionAsh5357 (approved):

Issue #132, opened 2024-07-07
Add enchantment documentation


IchHabeHunger54:

ChampionAsh5357:
Closed by #165

PR #165, opened 2024-09-17
Add Enchantments page

hjake123:
Adds a page under Resources > Server that describes how to make custom Enchantments and Enchantment Component Types in 1.21.x, as well as a small amount about how to work with the latter. 

The relevant _minecraft.wiki_ pages are frequently referenced and linked to, but the written content is my own.

Looking for feedback on this draft! 

This should address issue #132 

------------------
Preview URL: https://pr-165.neoforged-docs-previews.pages.dev


hjake123:
Should have mentioned that this is to address issue #132

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
+# Enchantments
+
+Enchantments are enchancements that can be applied to tools and other items. As of 1.21, enchantments are stored on items as [Data Components], are defined in JSON, and are comprised of Enchantment Components.
```
ChampionAsh5357, line 3: I don't like the word 'enhance' here since it's effects may be detrimental or nothing at all depending on how the user chooses to implement them.

ChampionAsh5357, line 3: Enchantment effect components.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+
+A new enchantment can be added by creating a JSON file in your namespace's `enchantment` datapack subfolder. For example, to create an enchantment called `examplemod:example_enchant`, one could create the file `data/examplemod/enchantment/example_enchantment.json`. Enchantments added this way are automatically registered into the game as part of the data pack loading process.
```
ChampionAsh5357, line 5: Enchantments can only be added in this manner. It would probably be best to rephrase that last sentence.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+
+The JSON format for Enchantments is summarized on the [Enchantment definition Minecraft wiki page].
```
ChampionAsh5357, line 7: This should also be properly explained here.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Enchantment Components
+Enchantment Components are specially-registered [Data Components] that determine what an enchantment does. Enchantment Component Types must be [registered] to `BuiltInRegistries.ENCHANTMENT_EFFECT_COMPONENT_TYPE`.
+:::note
+There are many already existing Enchantment Component Types in vanilla that can be used to create enchantment logic.
+:::
```
ChampionAsh5357, line 13: What are they? How can they be used?

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+
+For simple enchantments that act as a marker for custom logic, `DataComponentType<Unit>` may be sufficient. Such a component type could be registered, for example, using a DeferredRegister:
```
ChampionAsh5357, line 15: This is not a good way to say this. The point of the enchantment effect components is to store the data of the effects to apply when put onto an object. Generally, your custom logic will likely have some configurable data, which is what would be the generic instance and not just unit.

Additionally, your logic may require certain conditions to work, which may want to have the effect wrapped in a `ConditionalEffect`.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<Unit>> EXAMPLE =
+    ENCHANTMENT_COMPONENT_TYPES.register("example",
+        () -> DataComponentType.<Unit>builder()
+            .persistent(Unit.CODEC)
+            .build());
+```
+
+### Testing For Enchantment Components
+To check if an enchantment component is present on a given item stack, one must first get the item's `minecraft:enchantment` component. The value of this component is an `ItemEnchantments` instance containing a set of all the Enchantments on the item, which can be acquired using `ItemEnchantments#keySet()`. The enchantment components of a given enchantment can then be acquired using the `Enchantment::effects()` method:
```
ChampionAsh5357, line 28: This should be done with either `runIterationOnItem`, `runIterationOnEquipement`, `pickHighestLevel`, `has`, or `hasTag`. These more or less handle what you are trying to write, which would generally be more limiting as most enchantments are supposed to return some value based on the iterable run.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+            // Do something!
+    }
+}
+```
+
+## Specialized Enchantment Component Types
+Vanilla uses a few specific classes when building its enchantment component types to achieve the customizability present in the vanilla enchantment components.
+
+### `ConditionalEffect`
+Wrapping the type in `List<ConditionalEffect<?>>` permits the use of [predicates] when defining enchantments. This allows some effects to activate situationally. Most vanilla enchantments use this.
```
ChampionAsh5357, line 43: You do not need the list. The list is if you want multiple conditional effects. The conditional effect on its own is good enough unless you want multiple effects to apply.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+Registering such an effect can be done as follows:
+```java
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<List<ConditionalEffect<Unit>>>> EXAMPLE_CONDITIONAL_EFFECT =
+    ENCHANTMENT_COMPONENT_TYPES.register("example_conditional",
+        () -> DataComponentType.<List<ConditionalEffect<Unit>>>builder()
```
ChampionAsh5357, line 48: This should not be a unit with a list as there would be no point to the list. There's only one state the conditional can be in: true or false.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+            .build());
+```
+The parameters to `ConditionalEffect.codec` are the codec for the template type of the `ConditionalEffect<T>`, followed by some `LootContextParamSets` item. There are many of these to choose from, so choose carefully.
+
+### Value Effects
+[Value Effects] are used for enchantments that change some numerical value to a different degree depending on the enchantment's level, and are implemented by the class `EnchantmentValueEffect`. Enchantments like Knockback, Looting, an Sharpness use this kind of component.
+
+A vanilla-conforming way to adjust values based on custom Value Effect components is to invoke one of the overloads of `EnchantmentHelper#runIterationOnItem` and pass in a lambda that alters a `MutableFloat` or similar mutable object based on the results:
+```java
+MutableFloat mutable_value = new MutableFloat(unmodified_value);
```
ChampionAsh5357, line 59: Camel case and explain why you are using a mutable object here.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+    (value_effect) -> mutable_value.setValue(value_effect.process(enchant_level, server.random, mutable_value.getValue()))
+));
+```
+
+Custom numerical operations for use in Value Enchantment blocks can be added by registering a subclass of `EnchantmentValueEffect` through `BuiltInRegistries.ENCHANTMENT_VALUE_EFFECT_TYPE`.
+
+### Entity Effects
+[Entity Effects] are used for enchantments that cause special effects to either the wielder or (if applicable) the target of an attack when the enchanted item is used to attack another entity. Enchantments like Flame, Fire Aspect, and Channeling use this kind of component.
+
+Custom `EnchantmentEntityEffect` extensions can be registered through `BuiltInRegistries.ENCHANTMENT_ENTITY_EFFECT_TYPE`.
```
ChampionAsh5357, line 72: This is worded in a misleading way. The enchantment entity effect is just a data object that can be a part of an enchantment effect component. For example, there are 'entity effects' for after the entity finishes an attack or hits a block.

It's a separate interface to make it easier to implement how that effect is applied to the entity, rather than writing it out in its place of usage, since it could be used in multiple places.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Location Based Effects
+[Location Based Effects] are used for effects that need to reference specific places in the world relative to the player. Frost Walker is a prime example of a Location Based Effect, and Entity Effects are a subclass of Location Based Effects.
```
ChampionAsh5357, line 75: Relative to an entity, but this the same sort of issue with the entity effects section that it's sort of worded misleadingly. These are the component objects.

ChampionAsh5357 (changes requested):
Thank you for the contribution!

Generally, this needs a lot more detail. It is too sparse to have an understanding on how to properly use the system. One example of this is the value effects, as that is hard to wrap your head around if you do not have a solid understanding of what is going on.

You should incorporate more information and examples of how one would implement these custom methods. Create a fake object that holds some arbitrary conditions. It should help the reader better understand how to create an enchantment.

docs/resources/server/enchantments.md
hjake123: By this, do you mean I should copy/paraphrase the JSON format section from the Minecraft wiki?

docs/resources/server/enchantments.md
hjake123: I... did not know `EnchantmentHelper#has` existed.

docs/resources/server/enchantments.md
hjake123: Alright

hjake123: Right, that is their name isn't it... I've made this adjustment everywhere.

docs/resources/server/enchantments.md
ChampionAsh5357: More or less. It shouldn't be a straight copy and should have more details related to how it works from a modder's perspective rather than a modpack makers'. It would also be good to add a datagen equivalent, similar to what's done in the biome modifiers PR.

hjake123: I've added the JSON schema, but am not sure about how datagen works for enchantments.

ChampionAsh5357: It's the same way as any other datapack registry. You are simply creating the object within a registry set builder like so: https://docs.neoforged.net/docs/worldgen/biomemodifier#datagenning-biome-modifiers

docs/resources/server/enchantments.md
hjake123: Is it important to replicate this information here from the Minecraft Wiki?

ChampionAsh5357: They should be written for the intended audience of a modder. While it may have similar information to the Minecraft wiki, it'll probably be a bit more descriptive here, especially since this should explain how to create your own component types.

hjake123:
I think I've applied all these changes, let me know if there's more still to do!

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+            "per_level_above_first": 0.5
+          }
+        }
+      }
+    ]
+  }
+```
+
+#### Using Enchantment Value Effect Components
+One way to adjust values based on custom Value Effect components is to invoke one of the overloads of `EnchantmentHelper#runIterationOnItem`. This function accepts an `EnchantmentHelper.EnchantmentVisitor`, which is a functional interface that accepts an enchantment and its level, and is invoked on all of the enchantments that the given itemstack has. While any consumer of those two values will work, the `Enchantment` class provides a handy function that fits this interface (and is used often by vanilla) -- `Enchantment#applyEffects`. This function takes a `List<ConditionalEffect<EnchantmentValueEffect>>`, a `LootContext`, and a function that consumes a value effect. The function verifies that all of the `ConditionalEffect`s are satisfied given the provided `LootContext`, and if so, invokes the provided function.
```
ChampionAsh5357, line 101: I feel like this a bit too brief to understand what is occurring. Adjusting the value is an iterative process where the the mutable value is continually updated by each successive enchantment effect applied. This means that multipliers have the potential to grow exponentially depending on the circumstance.

There needs to be more detail to understand what's going on, and how it compounds or relates to the final returned value.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+
+To actually perform the adjustment, use `EnchantmentValueEffect#process`, which takes the enchantment level, a random value (in case the binomial_random Level Based Value was asked for), and a float, then returns the adjusted float based on the settings provided to the `EnchantmentValueEffect` instance. A convenient trick is to invoke this method inside a lambda argument to the aforementioned `Enchantment#applyEffects` method, and to have it update a `MutableFloat` with the value to allow it to escape the lambda and be put to use elsewhere.
+
+```java
+MutableFloat mutableValue = new MutableFloat(unmodified_value);
```
ChampionAsh5357, line 106:
```suggestion
MutableFloat mutableValue = new MutableFloat(unmodifiedValue);
```

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+- `minecraft:projectile_spread`: Modifies the maximum spread of projectiles in degrees from the direction they were fired. Used by Multishot.
+- `minecraft:trident_return_acceleration`: Causes the trident to return to its owner, and modifies the acceleration applied to this trident while doing so. Used by Loyalty.
+
+Other:
+- `minecraft:block_experience`: Modifies the amount of XP from breaking a block. Used by Silk Touch.
+- `minecraft:fishing_time_reduction`: Reduces the time it takes for the bobber to sink while fishing with this rod by the given number of seconds. Used by Lure.
+- `minecraft:fishing_luck_bonus`: Modifies the amount of [luck] used in the fishing loot table. Used by Luck of the Sea.
+
+#### Defined as `DataComponentType<List<TargetedConditionalEffect<EnchantmentValueEffect>>>`
+- `minecraft:equipmentment_drops`: Modifies the chance of equipment dropping from an entity killed by this weapon. Used by Looting.
```
ChampionAsh5357, line 149: Examples? Though they should probably be in a separate page(s) to avoid clutter.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+- `minecraft:replace_disk`: Replaces a disk of blocks.
+- `minecraft:run_function`: Runs a specified [datapack funcion].
+- `minecraft:set_block_properies`: Modifies the block state properties of the specified block.
+- `minecraft:spawn_particles`: Spawns a particle.
+- `minecraft:summon_entity`: Summons an entity.
+
+For more detail on each of these, please look at the [relevant minecraft wiki page].
+
+### Location Based Effects
+[Location Based Effect Components] are like Entity Effect Components, but they instead contain Location Based Effects. These are used for effects that need to reference specific places in the world relative to the wielder. Entity Effects are a subclass of Location Based Effects.
```
ChampionAsh5357, line 209: These should come first as entity effects are a subtype. Also, restating the title of the section does not explain what location-based effects are.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+- `minecraft:damage_item`: Damages this item's durability.
+- `minecraft:explode`: Summons an explosion. 
+- `minecraft:ignite`: Sets the entity on fire.
+- `minecraft:play_sound`: Plays a specified sound.
+- `minecraft:replace_block`: Replaces a block at a given offset.
+- `minecraft:replace_disk`: Replaces a disk of blocks.
+- `minecraft:run_function`: Runs a specified [datapack funcion].
+- `minecraft:set_block_properies`: Modifies the block state properties of the specified block.
+- `minecraft:spawn_particles`: Spawns a particle.
+- `minecraft:summon_entity`: Summons an entity.
```
ChampionAsh5357, line 204: Aren't these both entity effects and location based effects? Attributes is the only difference between the two from what I understand.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+    "base": 3,
+    "per_level_above_first": 1
+  },
+  "slots": [
+    "any"
+  ],
+  "supported_items": <supported item list>,
+  "weight": 30
+}
+```
```
ChampionAsh5357, line 300: See biome modifiers on how to use tabs for this.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+                    .build()
+        )
+    )
+);
+
+```
+
+This will produce the following JSON data when the data generator runs:
+
+```json
```
ChampionAsh5357, line 278:
```suggestion
```json5
```

ChampionAsh5357 (changes requested):
Good work! I will say there needs to be a lot more detail and narrative dialogue to help with the understanding of the document. It was hard to get a sense of how the system worked without already being aware of the logic itself. It's best if you take a step back and imagine if someone who had no knowledge of the system, would they be able to understand what you've written.

docs/resources/server/enchantments.md
hjake123: Ah, that's true. I was confused by the apparent duplication between entity and location based effects when I wrote this I think

docs/resources/server/enchantments.md
hjake123: I think I got it!

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<ExampleEffectData>> EXAMPLE =
+    ENCHANTMENT_COMPONENT_TYPES.register("example",
+        () -> DataComponentType.<ExampleEffectData>builder()
+            .persistent(ExampleEffectData.CODEC)
+            .build());
+```
+
+### `ConditionalEffect`
+Wrapping the type in `ConditionalEffect<?>` permits the use of [predicates] when defining enchantments. This allows some effects to activate situationally.
```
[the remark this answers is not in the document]
hjake123: I think I've done that somewhat but I do still have a few cases where I use invented names that resemble the codebase names to prevent there being too many code reference blocks. For example, calling a section Value Effect Components instead of `EnchantmentValueEffect`s. 

Should I change the section headers to use class names?

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+```java
+// In some registration class
+public static final DeferredRegister<DataComponentType<?>> ENCHANTMENT_COMPONENT_TYPES = DeferredRegister.create(BuiltInRegistries.ENCHANTMENT_EFFECT_COMPONENT_TYPE, "examplemod");
+
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<ExampleEffectData>> EXAMPLE =
+    ENCHANTMENT_COMPONENT_TYPES.register("example",
+        () -> DataComponentType.<ExampleEffectData>builder()
+            .persistent(ExampleEffectData.CODEC)
+            .build());
+```
```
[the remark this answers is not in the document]
hjake123: I've decided to split the built-in component types into their own page, following the example of the Recipes and Data Maps sections. I think this should help the main Enchantments page flow better, since it doesn't have to be padded out by the list of already existing components.

docs/resources/server/enchantments.md
```diff
@@ -0,0 +1,319 @@
...
+```java
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<ConditionalEffect<Unit>>> EXAMPLE_CONDITIONAL_EFFECT =
+    ENCHANTMENT_COMPONENT_TYPES.register("example_conditional",
+        () -> DataComponentType.ConditionalEffect<Unit>>builder()
+            .persistent(ConditionalEffect.codec(Unit.CODEC, LootContextParamSets.EMPTY))
+            .build());
+```
+The parameters to `ConditionalEffect.codec` are the codec for the template type of the `ConditionalEffect<T>`, followed by some `LootContextParamSets` item. There are many of these to choose from, so choose carefully.
+
+Whether a given item stack has a specific Enchantment Effect Component can be tested for with `EnchantmentHelper#has`.
```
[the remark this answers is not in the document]
hjake123: Ah, it isn't, that was a leftover from reordering things.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,287 @@
...
+}
+```
+
+## Enchantment Effect Components
+Enchantment Effect Components are specially-registered [Data Components] that determine how an enchantment functions. The type of the component defines its effect, while the data it contains is used to inform or modify that effect. For instance, the `minecraft:damage` component modifies the damage that a weapon deals by an amount determined by its data.
+
+Vanilla defines various [built-in enchantment effect components], which are used to implement all vanilla enchantments.
+
+### Custom Enchantment Effect Components
+Enchantment Effect Component Types must be [registered] to `BuiltInRegistries.ENCHANTMENT_EFFECT_COMPONENT_TYPE`, which takes a `DataComponentType<?>`. For example, one could register an Enchantment Effect Component that can store an `ExampleEffectData` object as follows:
```
IchHabeHunger54, line 92:
```suggestion
Enchantment Effect Component Types must be [registered] to `BuiltInRegistries.ENCHANTMENT_EFFECT_COMPONENT_TYPE`, which takes a `DataComponentType<?>`. For example, you could register an Enchantment Effect Component that can store an `ExampleEffectData` object as follows:
```

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,287 @@
...
+public static final DeferredRegister<DataComponentType<?>> ENCHANTMENT_COMPONENT_TYPES = DeferredRegister.create(BuiltInRegistries.ENCHANTMENT_EFFECT_COMPONENT_TYPE, "examplemod");
+
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<ExampleEffectData>> EXAMPLE =
+    ENCHANTMENT_COMPONENT_TYPES.register("example",
+        () -> DataComponentType.<ExampleEffectData>builder()
+            .persistent(ExampleEffectData.CODEC)
+            .build());
+```
+
+There are no inheritance requirements on the data held by an Enchantment Effect Component, but it may be helpful to refer to vanilla to ensure compatibility with the vanilla helper methods.
```
IchHabeHunger54, line 105: Maybe show an example, even if there's no real requirements.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,287 @@
...
+
+### `ConditionalEffect`
+Wrapping the type in `ConditionalEffect<?>` allows the Enchantment Effect Component to take effect based on conditions informed by a `LootContext`. 
+
+Specifically, each `ConditionalEffect` contains another effect component, along with an `Optional<LootItemCondition>`. Since `LootItemContext` is a `Predicate<LootContext>`, it can be tested against a specified `LootContext` using `LootItemContext#test`.
+
+`ConditionalEffect` wraps this behavior, allowing one to simply call `ConditionalEffect#matches(LootContext context)` to determine if the effect should be allowed to run.
+
+Vanilla adds an additional helper method to further streamline the process of checking these conditions: `Enchantment#applyEffects()`. This method takes a `List<ConditionalEffect<T>>`, evaluates the conditions, and runs a `Consumer<T>` on each `T` contained by a `ConditionalEffect` whose condition was met. Since many of Vanilla Enchantment Effect Components are defined as `List<ConditionalEffect<?>>`, these can be directly plugged into the helper method like so:
```
IchHabeHunger54, line 114:
```suggestion
Vanilla adds an additional helper method to further streamline the process of checking these conditions: `Enchantment#applyEffects()`. This method takes a `List<ConditionalEffect<T>>`, evaluates the conditions, and runs a `Consumer<T>` on each `T` contained by a `ConditionalEffect` whose condition was met. Since many vanilla enchantment effect components are defined as a `List<ConditionalEffect<?>>`, these can be directly plugged into the helper method like so:
```

IchHabeHunger54 (changes requested):
Thank you for the contribution! I left a few remarks, mostly regarding grammar, typos or formatting; content-wise this already looks very good.

A few general requests:
- Do not capitalize nouns like "Enchantment Effect Components" outside of headings.
- Add a blank line after every subheading, for consistency with the rest of the docs.
- Add a blank line before and after a codeblock.
- Format JSON blocks to 2 space indents.

And as a final positive remark, I really like the usage of tabs for the datagen section. We should probably adapt that to other parts of the docs.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,183 @@
+# Built-In Enchantment Effect Components
+
+Vanilla Minecraft provides numerous different types of Enchantment Effect Components for use in [enchantment] definitions. This article will explain each, including their usage and in-code definition.
+
+## Value Effect Components
+[Value Effect Components] are used for enchantments that alter a numerical value somewhere in the game, and are implemented by the class `EnchantmentValueEffect`.
```
IchHabeHunger54, line 6: Consider replacing the inline links with something like this:
```
_See also: [Value Effect Components] on the Minecraft Wiki_
```
Similar for the other links below.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,183 @@
...
+The [Attribute Effect Component], `minecraft:attributes`, is a unique Enchantment Entity Effect Component of type `EnchantmentAttributeEffect` that is not registered as a Location Based Effect Component. It is used to apply attribute modifiers to the wielder of the enchantment, which are then removed when the enchanted item is no longer equipped. The JSON format is as follows:
+```json5
+"minecraft:attributes": [
+    {
+        "amount": {
+            "type": "minecraft:linear",
+            "base": 1,
+            "per_level_above_first": 1
+        },
+        "attribute": "namespace:class.attribute_name",
```
IchHabeHunger54, line 156:
```suggestion
        "attribute": "namespace:attribute_name",
```
The `class.` qualifier adds more confusion imo, Mojang keeping this qualifier is already confusing enough.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,287 @@
...
+    // This line actually performs the value adjustment.
+    // Each time it runs, mutableValue is set to (exampleData's value) + mutableValue + enchantLevel.
+));
+
+// Use mutableValue elsewhere in your game logic.
+```
+</TabItem>
+
+<TabItem value="explainer" label="Explanation">
+Consider an Enchantment Effect Component called `INCREMENT`, defined as `List<ConditionalEffect<Increment>>`. Let the `Increment` object it contains be a wrapper around an integer that defined a method `add(int x)`, which adds its internal value to the provided integer and returns the result. Imagine that you want to use this object to increase the count of another integer value within your item's `use` method -- say, to increase the number of times it performs some repeated effect.
```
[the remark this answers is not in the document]
hjake123: I used academic tone here in an attempt to avoid using "you" and "we" pronouns at any cost, but it's good to know I can use them here.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,287 @@
...
+    // Bounded by [1, 255].
+
+    "max_cost": {
+        "base": 45,
+        "per_level_above_first": 9
+    },
+    // The maximum cost of this enchantment in levels. 
+    // The actual cost during a particular enchantment session will be randomized between this and `min_cost`.
+    // Blocks that override IBlockExtension#getEnchantPowerBonus() add their bonus to the player's levels to determine if the player meets an enchantment's cost.
+    // Therefore, costs above 30 require the use of bookshelves or other enchantment boosting blocks to be obtainable in survival.
```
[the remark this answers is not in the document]
hjake123: I think I'll need to do more research here, the mechanics are pretty dense.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,183 @@
...
+- `minecraft:projectile_spread`: Modifies the maximum spread of projectiles in degrees from the direction they were fired. Used by Multishot.
+- `minecraft:trident_return_acceleration`: Causes the trident to return to its owner, and modifies the acceleration applied to this trident while doing so. Used by Loyalty.
+
+Other:
+- `minecraft:block_experience`: Modifies the amount of XP from breaking a block. Used by Silk Touch.
+- `minecraft:fishing_time_reduction`: Reduces the time it takes for the bobber to sink while fishing with this rod by the given number of seconds. Used by Lure.
+- `minecraft:fishing_luck_bonus`: Modifies the amount of [luck] used in the fishing loot table. Used by Luck of the Sea.
+
+#### Defined as `DataComponentType<List<TargetedConditionalEffect<EnchantmentValueEffect>>>`
+- `minecraft:equipmentment_drops`: Modifies the chance of equipment dropping from an entity killed by this weapon. Used by Looting.
```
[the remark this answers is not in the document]
hjake123: Mine...

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,287 @@
...
+```json5
+{
+    "description": {
+        "translate": "enchantment.examplemod."
+    },
+    // The text component that will be used as the in-game name of the enchantment.
+    // Can be a translation key or a literal string. 
+    
+    "supported_items": "#examplemod:enchantable/example",
+    // The items or item tag that this enchantment can be applied to.
```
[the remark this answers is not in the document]
hjake123: Is something like this what you were looking for?
```
 // Which items this enchantment can be applied to.
 // Can be either an item id, such as "minecraft:trident",
 // or a list of item ids, such as ["examplemod:red_sword", "examplemod:blue_sword"]
 // or an item tag, such as "#examplemod:enchantable/enchant_name".
 // Note that this doesn't cause the enchantment to appear for these items in the enchanting table.
```

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,183 @@
...
+            "base": 1.0,
+            "per_level_above_first": 0.5
+          }
+        }
+      }
+    ]
+  }
+```
+
+The object within the `value` block is a [LevelBasedValue], which can be used to have a Value Effect Component that changes the intensity of its effect by level.
```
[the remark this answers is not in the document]
hjake123: It mentions them but I'm not sure there's a spot where that article defines them. I guess the bullet list in the Number Provider section? I've put the link to there for now but I'm not sure...

docs/resources/server/enchantments/index.md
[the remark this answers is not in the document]
hjake123: Think I've got it.

docs/resources/server/enchantments/index.md
hjake123: Since I lead straight into a section where I'm discussing one of the most important of these interactions (the one with ConditionalEffect), and the others would require discussing classes that aren't mentioned outside the other entry, I think I might just cut this.

docs/resources/server/enchantments.md
hjake123: Do you think it would be good enough to refer readers to the vanilla datapack for usage examples?

docs/resources/server/enchantments/builtin.md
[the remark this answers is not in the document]
IchHabeHunger54: Yes, that's the part I meant.

docs/resources/server/enchantments/index.md
IchHabeHunger54: Works for me too.

docs/resources/server/enchantments.md
hjake123: The Sharpness example is now very close to this section so I think there is enough example of using these. If you disagree please reopen this!

IchHabeHunger54 (dismissed):
Didn't go through it in detail again, but looks good overall. Approved by me.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+}
+```
+
+### Enchantment Costs
+
+The `max_cost` and `min_cost` fields specify boundaries for how much enchanting power is needed to create this enchantment. There is a somewhat convoluted procedure to actually make use of these values, however.
+
+First, the table takes into account the return value of `IBlockStateExtension#getEnchantPowerBonus()` for the surrounding blocks. From this, it derives a 'base cost' for each slot. This cost is shown in-game as the green numbers besides the enchantments in the menu. For each enchantment, the base cost is modified twice by a random value derived from the item's enchantability (its return value from `IItemStackExtension#getEnchantmentValue()`), like so:
+
+`(Modified Cost) = (Base Cost) + random.nextInt(e / 4 + 1) + random.nextInt(e / 4 + 1)`, where `e` is the enchantability score.
```
ChampionAsh5357, line 102: I think they way you are wording this is wrong. The costs are calculated from `EnchantmentHelper#getEnchantmentCost`. Using those costs, a random enchantment is selected from an available enchantment list that can be obtained for that cost via `EnchantmentHelper#selectEnchantment`.

What you're describing here is how the enchantment level is chosen, not the cost.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+### Custom Enchantment Effect Components
+
+Enchantment effect component types must be [registered] to `BuiltInRegistries.ENCHANTMENT_EFFECT_COMPONENT_TYPE`, which takes a `DataComponentType<?>`. For example, you could register an enchantment effect component that can store an `ExampleEffectData` object as follows:
+
+```java
+// In some registration class
+public static final DeferredRegister<DataComponentType<?>> ENCHANTMENT_COMPONENT_TYPES = DeferredRegister.create(BuiltInRegistries.ENCHANTMENT_EFFECT_COMPONENT_TYPE, "examplemod");
+
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<ExampleEffectData>>> EXAMPLE =
```
ChampionAsh5357, line 122: And what is `ExampleEffectData`? This does not explain how a custom effect is made. Start with explaining that a custom enchantment effect component is implemented entirely by the user. The only thing that specifically needs to be registered is the data component that is stored within the enchantment itself. From there, the user must then search the specified object for enchantments and check if there enchantment is there. Then, implement the logic. Just creating the custom component does nothing.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+    ENCHANTMENT_COMPONENT_TYPES.register("example",
+        () -> DataComponentType.<ExampleEffectData>builder()
+            .persistent(ExampleEffectData.CODEC)
+            .build());
+```
+
+There are no inheritance requirements on the data held by an enchantment effect component.
+
+### `ConditionalEffect`
+Wrapping the type in `ConditionalEffect<?>` allows the enchantment effect component to take effect based on conditions informed by a [LootContext].
```
ChampionAsh5357, line 132: I don't understand this sentence. I understand you're trying to say the `LootItemCondition` is tested with the provided `LootContext` the enchanted object is used in, but it is worded strangely as it's not 'informed' but rather 'given' the context to work within.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+            .persistent(ExampleEffectData.CODEC)
+            .build());
+```
+
+There are no inheritance requirements on the data held by an enchantment effect component.
+
+### `ConditionalEffect`
+Wrapping the type in `ConditionalEffect<?>` allows the enchantment effect component to take effect based on conditions informed by a [LootContext]. 
+
+Specifically, each `ConditionalEffect` contains another effect component, along with an `Optional<LootItemCondition>`. Since `LootItemContext` is a `Predicate<LootContext>`, it can be tested against a specified `LootContext` using `LootItemContext#test`.
```
ChampionAsh5357, line 134: This is not how you would say it. Treat the `LootItemCondition` as if it were a predicate and not because it implements the `Predicate` functional interface.

Also `LootItemContext` -> `LootItemCondition`

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+`ConditionalEffect` wraps this behavior, allowing one to simply call `ConditionalEffect#matches(LootContext context)` to determine if the effect should be allowed to run.
```
ChampionAsh5357, line 136: Wraps what behavior? The testing mechanism? Because it acts more like a delegate holder where the method within the `ConditionalEffect` is called first before getting the effect method.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+Registering a custom `ConditionalEffect`-wrapped enchantment effect component type can be done as follows:
+
+```java
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<ConditionalEffect<ExampleData>>> EXAMPLE_CONDITIONAL_EFFECT =
+    ENCHANTMENT_COMPONENT_TYPES.register("example_conditional",
+        () -> DataComponentType.ConditionalEffect<ExampleData>builder()
+            .persistent(ConditionalEffect.codec(Unit.CODEC, LootContextParamSets.EMPTY))
+            .build());
```
ChampionAsh5357, line 157: The example doesn't line up... I honestly think you should just create a dummy object to reference. You don't need to implement the codec methods, but you should show that this is where it's coming from rather than referencing `ExampleData`, then passing in a `Unit#CODEC`, which is not `ExampleData`.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+The parameters to `ConditionalEffect.codec` are the codec for the generic `ConditionalEffect<T>`, followed by some `LootContextParamSets` entry.
+
+### Using Enchantment Effect Components
+
+Here is a full example using vanilla helper methods to work with a custom enchantment effect component.
+
+```java
+// Define an example data-bearing record.
+public record Increment(int value) {
+    public static final Codec<BoundEntity> CODEC = RecordCodecBuilder.create(instance ->
```
ChampionAsh5357, line 168:
```suggestion
    public static final Codec<Increment> CODEC = RecordCodecBuilder.create(instance ->
```

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+            instance.group(
+                    Codec.INT.fieldOf("value").forGetter(Increment::value),
+            ).apply(instance, Increment::new)
+    );
+
+    public int add(int x) {
+        return value() + x;
+    }
+}
```
ChampionAsh5357, line 177: See, this would be an example object you could show above and use for the rest of the base examples.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+```java
```
ChampionAsh5357, line 180: These do not need to be separate blocks. You can just say something like `// In some registry object class`

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+            .build());
+```
+
+```java
+// Somewhere in game logic where an `itemStack` is available.
+// `unmodifiedValue` is an integer.
+MutableInt mutableValue = new MutableInt(unmodifiedValue);
+EnchantmentHelper.runIterationOnItem(itemStack, (enchantmentHolder, enchantLevel) -> Enchantment.applyEffects(
+    // Isolates the ConditionalEffect<Increment> from the provided holder and wraps it in a list for applyEffects.
+    enchantmentHolder.value().getEffects(INCREMENT.value()),
```
ChampionAsh5357, line 195:
```suggestion
    enchantmentHolder.value().getEffects(INCREMENT.get()),
```

`Holder` would return a `DataComponentType<?>`, which means it would not line up with the generic, causing a compile error unless cast.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+    // Produces a LootContext. Other context helpers from the Enchantment class
+    // include itemContext, locationContext, entityContext, and blockHitContext.
+    Enchantment.damageContext(server, enchantLevel, target, damageSource),
```
ChampionAsh5357, line 199: Ok and why...? Where did these undefined variables come from. An example should specify what it's using. I don't think this example explains what you're trying to do, and is basically just a value component, which you'll be talking about below.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+));
+
+// Use mutableValue elsewhere in your game logic.
+```
+
+We define an enchantment effect component called `INCREMENT` as `List<ConditionalEffect<Increment>>`. The `Increment` object it contains is a wrapper around an integer that defines a method `add(int x)`, which adds its internal value to the provided integer and returns the result. Imagine that you want to use this object to increase the count of another integer value within your item's `use` method -- say, to increase the number of times it performs some repeated effect.
+
+First, invoke one of the overloads of `EnchantmentHelper#runIterationOnItem`. This function accepts an `EnchantmentHelper.EnchantmentVisitor`, which is a functional interface that accepts an enchantment and its level, and is invoked on all of the enchantments that the given itemstack has (essentially a `BiConsumer<Enchantment, Integer>`). While any consumer of those two values will work, the `Enchantment` class provides a handy function that fits this interface (and is used often by vanilla) -- `Enchantment#applyEffects`. This method is used as above to test the conditions of the `ConditionalEffect`s.
```
ChampionAsh5357, line 213: Mentions apply effects but does not use it?

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+
+To actually perform the adjustment, use the provided `Increment#add` method.
+
+Note that in this example, the level of the enchantment does not affect the outcome. This can be changed by using `enchantLevel` somewhere in the `Consumer<T>` lambda expression (the last line with code in the example). Any other information stored in the `ItemStack` can also be accessed from here, so other Data Components could be used to inform how the adjustment goes.
```
ChampionAsh5357, line 217: This is not true, as the condition that would be applied could use the enchantment level. You are simply not using it within the actual enchantment effect object.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,287 @@
...
+public static RegistrySetBuilder BUILDER = new RegistrySetBuilder();
+
+// Define the ResourceKey for our enchantment.
+static ResourceKey<Enchantment> EXAMPLE_ENCHANTMENT_KEY = ResourceKey.create(
+        Registries.ENCHANTMENT,
+        ResourceLocation.fromNamespaceAndPath("examplemod", "example_enchantment")
+);
+
+// Specify the enchantment definition of for our enchantment. 
+static Enchantment.EnchantmentDefinition EXAMPLE_ENCHANTMENT_DEFINITION = new Enchantment.EnchantmentDefinition(
```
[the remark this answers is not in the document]
ChampionAsh5357: It should be inlined as the definition can take in a `HolderSet`, so it can reference the bootstraps tag getter. Sure, you can static if you understand what you're doing, but the readers don't, so they should be provided the method in a way to avoid headaches.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
...
+  "min_cost": {
+    "base": 3,
+    "per_level_above_first": 1
+  },
+  "slots": [
+    "any"
+  ],
+  "supported_items": <supported item list>,
+  "weight": 30
+}
```
ChampionAsh5357, line 291: Add comments here like above.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,217 @@
...
+        "type": "minecraft:add",
+        "value": {
+          "type": "minecraft:linear",
+          "base": 1.0,
+          "per_level_above_first": 0.5
+        }
+      }
+    }
+  ]
+}
```
ChampionAsh5357, line 34: Add comments for each part to give a better idea of what's going on. Also, show a datagen tab version.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,307 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
```
ChampionAsh5357, line 2: Must be an `mdx` file to use imports.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,217 @@
+# Built-In Enchantment Effect Components
+
+Vanilla Minecraft provides numerous different types of enchantment effect components for use in [enchantment] definitions. This article will explain each, including their usage and in-code definition.
+
+## Value Effect Components
+
+_See also [Value Effect Components] on the Minecraft Wiki_
+
+Value effect components are used for enchantments that alter a numerical value somewhere in the game, and are implemented by the class `EnchantmentValueEffect`.
```
ChampionAsh5357, line 9: I would just specifically state here that multiple enchantments will iterate off the same value when applied.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,217 @@
...
+Custom numerical operations for use in value effect component definition blocks can be added by registering a subclass of `EnchantmentValueEffect` through `BuiltInRegistries.ENCHANTMENT_VALUE_EFFECT_TYPE`.
+
+The `EnchantmentValueEffect#process` method can be used to adjust values based on the provided numerical operations, like so:
+
+```java
+// `valueEffect` is an EnchantmentValueEffect instance.
+// `enchantLevel` is an integer representing the level of the enchantment
+float baseValue = 1.0;
+float modifiedValue = valueEffect.process(enchantLevel, server.random, baseValue);
+```
```
ChampionAsh5357, line 48: Ok. So, builtin components here means only the builtin components. If we're going to talk about implementing your own, it should be in the main file.

This should only reference existing vanilla types to use.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### Defined as `DataComponentType<List<TargetedConditionalEffect<EnchantmentValueEffect>>>`
+
+- `minecraft:equipment_drops`: Modifies the chance of equipment dropping from an entity killed by this weapon. Used by Looting.
+
+## Location Based Effect Components
+
+_See also: [Location Based Effect Components] on the Minecraft Wiki_
+
+ Location based effect components are components that contain an `EnchantmentLocationBasedEffect`. These components define actions to take that need to know where in the level the wielder of the enchantment is. They operate using two major methods: `EnchantmentEntityEffect#onChangedBlock`, which is called when the enchanted item is equipped and when the wielder changes their `BlockPos`, and `onDeactivate`, which is called when the enchanted item is removed.
```
ChampionAsh5357, line 93:
```suggestion
 Location based effect components are components that implement `EnchantmentLocationBasedEffect`. These components define actions to take that need to know where in the level the wielder of the enchantment is. They operate using two major methods: `EnchantmentEntityEffect#onChangedBlock`, which is called when the enchanted item is equipped and when the wielder changes their `BlockPos`, and `onDeactivate`, which is called when the enchanted item is removed.
```

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,217 @@
...
+
+Custom `EnchantmentLocationBasedEffect` extensions can be registered through `BuiltInRegistries.ENCHANTMENT_LOCATION_BASED_EFFECT_TYPE`. 
+
+Vanilla adds the following location based events:
+
+- `minecraft:all_of`: Runs a list of entity effects in sequence.
+- `minecraft:apply_mob_effect`: Applies a [mob effect] to the affected mob.
```
ChampionAsh5357, line 100: Missing `minecraft:attribute`

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,217 @@
...
+
+#### Defined as `DataComponentType<List<ConditionalEffect<EnchantmentLocationBasedEffect>>>`
+
+- `minecraft:location_changed`: Runs a location based effect when the wielder's Block Position changes and when this item is equipped. Used by Frost Walker and Soul Speed.
+
+## Entity Effect Components
+
+_See also [Entity Effect Components] on the Minecraft Wiki._
+
+Entity effect components are components that contain an `EnchantmentEntityEffect`, an extension of `EnchantmentLocationBasedEffect`. These override `EnchantmentLocationBasedEffect#onChangedBlock` to run `EnchantmentEntityEffect#apply` instead; this `apply` method is also directly invoked somewhere else in the codebase depending on the specific type of the component.
```
ChampionAsh5357, line 123:
```suggestion
Entity effect components are components that implement `EnchantmentEntityEffect`, an subtype of `EnchantmentLocationBasedEffect`. These override `EnchantmentLocationBasedEffect#onChangedBlock` to run `EnchantmentEntityEffect#apply` instead; this `apply` method is also directly invoked somewhere else in the codebase depending on the specific type of the component.
```

ChampionAsh5357, line 123: `apply` is  more of a method to say apply directly to the entity itself rather than wait for the block position of the player to change.

docs/resources/server/enchantments/builtin.md
```diff
@@ -0,0 +1,217 @@
...
+- `minecraft:tick`: Runs an entity effect each tick. Used by Soul Speed.
+- `minecraft:projectile_spawned`: Runs an entity effect after a projectile entity has been spawned from a bow or crossbow. Used by Flame.
+
+#### Defined as `DataComponentType<List<TargetedConditionalEffect<EnchantmentEntityEffect>>>`
+
+- `minecraft:post_attack`: Runs an entity effect after an attack damages an entity. Used by Bane of Arthropods, Channeling, Fire Aspect, Thorns, and Wind Burst.
+
+For more detail on each of these, please look at the [relevant minecraft wiki page].
+
+### Attribute Effect Component
```
ChampionAsh5357, line 168: Attribute is not an `EnchantmentEntityEffect`. It is an `EnchantmentLocationBasedEffect`.

ChampionAsh5357 (changes requested):
Great progress! However, I do think there needs to be a bit more explanation with your current reorganization. There are also some minor inaccuracies here and there that need to be fixed.

docs/resources/server/enchantments/index.md
hjake123: Wait, how is the tab rendering working in the preview if I'm not properly importing it?

docs/resources/server/enchantments/index.md
hjake123: ParchmentMC calls the array of these values 'costs', which is where I got the idea to call them that, but I can see that it makes more sense to call these enchantment levels.

docs/resources/server/enchantments/index.md
hjake123: I missed that I was still using the Unit codec, sorry.

docs/resources/server/enchantments/builtin.md
hjake123: This isn't a complete example of a data generated file, so how would I show data generation for it? I could show datagen that makes the entire Sharpness enchantment, but then it would feel out of scope for this section, wouldn't it?

docs/resources/server/enchantments/index.md
ChampionAsh5357: It appears that all files are parsed as mdx regardless of file extension. I'll revisit this with the rest of the docs team and see what we decide to do

docs/resources/server/enchantments/builtin.md
hjake123: I could have sworn... thanks.

docs/resources/server/enchantments/builtin.md
ChampionAsh5357: You could just have a comment at the top that says something like `// Passed for 'effects' in an Enchantment` and then show the relevant object.

docs/resources/server/enchantments/index.md
hjake123: I'm not really sure what treating it as if it were a `Predicate` entails aside from telling the user they can `Predicate#test()` it, but I will attempt to reword this section.

(Also oops sorry)

ChampionAsh5357: To clarify, the wording suggests that because the `LootItemCondition` is a subtype of predicate, that it can be tested. Rather, it should be that the `LootItemCondition` is a predicate itself, which just happens to implement `Predicate<LootContext>`.

It's like when you talk about supplied instances, we don't say that because it extends `Supplier#get`, but rather the instance is obtained from the `get` method.

It's just a bit of wording to avoid having something seemed tied to a given type even though it could be changed. Though, this is very much an 'irk' comment from me, so I apologize.

docs/resources/server/enchantments/index.md
hjake123: This example does use `Enchantment.applyEffects`, but it was at the end of a line. Moved it to be easier to spot.

docs/resources/server/enchantments/index.md
hjake123: I've removed discussion of the `LootItemCondition#test` method, since to be honest I'm not sure it was even relevant.

docs/resources/server/enchantments/index.md
hjake123: Should I resolve this to try to get in before the freeze on Monday?

ChampionAsh5357: You're welcome to. I'll probably be making the change myself during the freeze depending on what we decide.

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,322 @@
...
+
+Now, we can implement some game logic that makes use of this component to alter an integer value:
+
+```java
+// Somewhere in game logic where an `itemStack` is available.
+// `INCREMENT` is the enchantment component type holder defined above.
+// `value` is an integer.
+AtomicInteger atomicValue = new AtomicInteger(value);
+
+EnchantmentHelper.runIterationOnItem(stack, (enchantmentHolder, enchant_level) -> {
```
ChampionAsh5357, line 154:
```suggestion
EnchantmentHelper.runIterationOnItem(stack, (enchantmentHolder, enchantLevel) -> {
```

docs/resources/server/enchantments/index.md
```diff
@@ -0,0 +1,322 @@
...
+);
+```
+
+Registering a custom `ConditionalEffect`-wrapped enchantment effect component type can be done as follows:
+
+```java
+public static final DeferredHolder<DataComponentType<?>, DataComponentType<ConditionalEffect<Increment>>> CONDITIONAL_INCREMENT =
+    ENCHANTMENT_COMPONENT_TYPES.register("conditional_increment",
+        () -> DataComponentType.ConditionalEffect<Increment>builder()
+            .persistent(ConditionalEffect.codec(Increment.CODEC, LootContextParamSets.EMPTY))
```
ChampionAsh5357, line 195: So, for examples, you should never use constants that will never be used by a modder. Some parameter is required for any condition, as all conditions require some context.

I would leave a comment above this line saying that the context provided should include whatever the enchantment should be used for. If you are not sure, use one of the existing `ENCHANTED_*` or as a guideline.

ChampionAsh5357 (changes requested):
Thank you for all the work you've done! Just two more changes: one syntax based and the other due to providing a understandable baseline.

Once these two are addressed, I don't believe I have any more comments for its current state.

hjake123:
Got it! Thank you so much for your patience and advice in leading me through this project.

ChampionAsh5357 (approved):

IchHabeHunger54 (approved):

PR #173, opened 2024-10-06
Fix Advancements page

TathanDev:
This PR fix two things in the [advancements](https://docs.neoforged.net/docs/resources/server/advancements/) page.

## Supplier<ExampleCriterionTrigger>#value()
In the docs, there is written that for the [`performExampleAction`](https://docs.neoforged.net/docs/resources/server/advancements/#calling-criterion-triggers), we need to use get our registry but with the  non-existent `value()` method  and it should be replaced with the `get()` method.

## Parameter name for [AdvancementGenerator](https://docs.neoforged.net/docs/resources/server/advancements/#data-generation) 
This is a minor change. To datagen an advancement, we need to create a class that implements `AdvancementProvider.AdvancementGenerator`. This interface have the `generate` method with theses paramaters `HolderLookup.Provider registries`, `Consumer<AdvancementHolder> saver`, `ExistingFileHelper existingFileHelper` but in the docs, the parameter `consumer` is named `saver`

## AdvancementRewards.Builder#addLootTable() require a ResourceKey
In the docs, to add a lootTable, it says that the function `AdvancementRewards.Builder#addLootTable()` require a ResourceLocation but it actually need a `ResourceKey<LootTable>`

### DIFF
--- a/docs/resources/server/advancements.md
+++ b/docs/resources/server/advancements.md
@@ -231,7 +231,7 @@ builder.rewards(
     // Alternatively, use addExperience() to add to an existing builder.
     AdvancementRewards.Builder.experience(100)
     // Alternatively, use loot() to create a new builder.
-    .addLootTable(ResourceLocation.fromNamespaceAndPath("minecraft", "chests/igloo"))
+    .addLootTable(ResourceKey.create(Registries.LOOT_TABLE, ResourceLocation.fromNamespaceAndPath("minecraft", "blocks/dirt")))
     // Alternatively, use recipe() to create a new builder.
     .addRecipe(ResourceLocation.fromNamespaceAndPath("minecraft", "iron_ingot"))
     // Alternatively, use function() to create a new builder.
--- a/docs/resources/server/advancements.md
+++ b/docs/resources/server/advancements.md
@@ -231,7 +231,7 @@ builder.rewards(
     // Alternatively, use addExperience() to add to an existing builder.
     AdvancementRewards.Builder.experience(100)
     // Alternatively, use loot() to create a new builder.
-    .addLootTable(ResourceKey.create(Registries.LOOT_TABLE, ResourceLocation.fromNamespaceAndPath("minecraft", "blocks/dirt")))
+    .addLootTable(ResourceKey.create(Registries.LOOT_TABLE, ResourceLocation.fromNamespaceAndPath("minecraft", "chests/igloo")))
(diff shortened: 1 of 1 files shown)


docs/resources/server/advancements.md
```diff
@@ -247,7 +247,7 @@ builder.requirements(AdvancementRequirements.allOf(List.of("pickup_dirt")));
 
 // Save the advancement to disk, using the given resource location. This returns an AdvancementHolder,
 // which may be stored in a variable and used as a parent by other advancement builders.
-builder.save(saver, ResourceLocation.fromNamespaceAndPath("examplemod", "example_advancement"), existingFileHelper);
+builder.save(consumer, ResourceLocation.fromNamespaceAndPath("examplemod", "example_advancement"), existingFileHelper);
```
ChampionAsh5357, line 250: I disagree with this change. `AdvancementGenerator#generator` passes a `Consumer<AdvancementHolder> saver`, which is the name of the parameter in NeoForge. I don't like consumer in general because it doesn't explain what the lambda is supposed to be doing.

docs/resources/server/advancements.md
```diff
@@ -231,7 +231,7 @@ builder.rewards(
     // Alternatively, use addExperience() to add to an existing builder.
     AdvancementRewards.Builder.experience(100)
     // Alternatively, use loot() to create a new builder.
-    .addLootTable(ResourceLocation.fromNamespaceAndPath("minecraft", "chests/igloo"))
+    .addLootTable(ResourceKey.create(Registries.LOOT_TABLE, ResourceLocation.fromNamespaceAndPath("minecraft", "blocks/dirt")))
```
ChampionAsh5357, line 234: Would you mind changing this on the 1.20.6 docs as well? I missed this there.

ChampionAsh5357 (changes requested):
Hello! Thanks for the contribution!

Basically just one thing to revert and one comment due to it being outdated in a previous version of the docs.

docs/resources/server/advancements.md
TathanDev: I didn't see a part in the 1.20.6 docs where the `.addLootTable()` method is mentionned

ChampionAsh5357: My apologies, I thought that the server resources rework was merged before 1.21. This can be resolved.

ChampionAsh5357 (approved):

PR #184, opened 2024-10-24
Add Feature Flags documentation

ApexModder:
Initial rough write documentation for feature flags.

Goes over how to register and lookup a custom flag, mark elements as requiring said flag and how provide a feature pack to allow users to enable said flag.

This depends on either of the existing flag prs https://github.com/neoforged/NeoForge/pull/1619 https://github.com/neoforged/NeoForge/pull/1538

First time writing up documentation, please let me know how/where I can improve anything.


thomasglasser:
It may be better for readability to move the external section below the built-in section

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+**Installing into a server**
+
+Enabling Feature Packs can only be done during initial world creation, and cannot be disabled once enabled.
+
+1. Create the directory `./world/datapacks`
+2. Upload the datapack zip file into the newly created directory
+3. Open your servers `server.properties` file
+4. Add the datapack zip file name (excluding `.zip`) to `initial-enabled-packs` (separating each pack by a `,`)
+
+The Feature Pack is now installed and enabled
```
IchHabeHunger54, line 202: This line seems unnecessary to me.

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+:::
+
+**Installing into a server**
+
+Enabling Feature Packs can only be done during initial world creation, and cannot be disabled once enabled.
+
+1. Create the directory `./world/datapacks`
+2. Upload the datapack zip file into the newly created directory
+3. Open your servers `server.properties` file
+4. Add the datapack zip file name (excluding `.zip`) to `initial-enabled-packs` (separating each pack by a `,`)
```
IchHabeHunger54, line 200: Maybe add an example here.

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+
+The Feature Pack is now installed and enabled
+
+**Built-In**
+
+Built-In packs are bundled with your mod and are made available to the game using the `AddPackFindersEvent` event.
```
IchHabeHunger54, line 206:
```suggestion
Built-in packs are bundled with your mod and are made available to the game using the `AddPackFindersEvent` event.
```

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+```java
+@SubscribeEvent
+public static void addFeaturePacks(final AddPackFindersEvent event) {
+    event.addPackFinders(
+            // 'packLocation': Path relative to your mods 'resources' pointing towards this pack
```
IchHabeHunger54, line 211: The parameter name should not be added to the comment, an explanation of what is expected here should be enough.

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+            // 'packType': What kind of resources are contained within this pack
+            // 'CLIENT_RESOURCES' for packs with client assets (resource packs)
+            // 'SERVER_DATA' for packs with server data (data packs)
+            PackType.SERVER_DATA,
+            
+            // 'packNameDisplay': Display name shown in the Experiments screen
+            Component.literal("ExampleMod: Experiments"),
+            
+            // 'packSource': In order for this pack to load and enable feature flags this *MUST* be 'FEATURE'
+            // any of 'PackSource' type is invalid here
```
IchHabeHunger54, line 223:
```suggestion
            // In order for this pack to load and enable feature flags, this MUST be 'FEATURE',
            // any other PackSource type is invalid here
```

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+            PackSource.FEATURE,
+            
+            // 'alwaysActive': If 'true' this pack is always active and cannot be disabled, should always be 'false' for feature packs
```
IchHabeHunger54, line 226:
```suggestion
            // If this is true, the pack is always active and cannot be disabled, should always be false for feature packs
```

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+            // Must be a valid registered flag
+            "examplemod:experimental"
+        ]
+    ]
+}
+```
+
+There are a couple methods in which you can provide a feature pack to your users.
+
+Installing these packs differs depending on if its being installed into a single player world or a server.
```
IchHabeHunger54, line 169: I don't really understand these sentences in context. Do you mean that there's different ways of installing (built-in vs. external, and client vs. server)? If so, state this here directly, for example like so:

"There are a couple of ways for users to obtain a feature pack, namely installing them from an external source as a datapack, or downloading a mod that has a built-in feature pack. Both of these then need to be installed differently depending on the physical side." Also link to the Physical Sides section in the Sides article.

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+
+**External**
+
+External packs are provided to your users in datapack form.
+
+**Installing into a single player world**
```
IchHabeHunger54, line 175:
```suggestion
#### Installation in Singleplayer
```
Make this a H4 since it is a subsection of External, which would be a H3. Also add a new line after the heading.

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+
+:::note
+External Feature Packs do not show up in the Experiments screen.
+
+To disable these packs after enabling them, navigate back into the datapacks screen and move the external packs back into `Available` from `Selected`.
+
+The Experiments screen only allows toggling on or off Built-In Feature Packs.
+:::
+
+**Installing into a server**
```
IchHabeHunger54, line 193:
```suggestion
#### Installation in Multiplayer
```

IchHabeHunger54 (changes requested):
Thank you for the contribution, and sorry for taking so long. This is looking quite good for a first version, the explanations are well done, and most change requests are of grammatical or structural nature.

A few things occurring across the entire documentation, which I'd ask you to change, are:

- The term "Feature Flags" is often capitalized. Terms like that should not be capitalized outside of headings. Otherwise we'd have to capitalize Block and Item every single time as well, which just looks weird.
- Please ensure that there is one blank line before and after each code block.
- Use H2s and H3s instead of H3s and bold paragraphs.

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
...
+    // This is used when your menu is not encoding complex data during `player.openMenu`
+    // (windowId, inventory) -> new ExperimentalMenu(windowId, inventory),
+
+    // Using NeoForge IContainerFactory
+    // This is used when you wish to read complex data encoded during `player.openMenu`
+    // Casting here is important as `MenuType` specificly wants `MenuSupplier`
+    (IContainerFactory<ExperimentalMenu>) (windowId, inventory, buffer) -> new ExperimentalMenu(windowId, inventory, buffer),
+    
+    // mark as requiring the 'EXPERIMENTAL' flag
+    FeatureFlagSet.of(EXPERIMENTAL)
```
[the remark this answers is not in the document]
ApexModder: Im confused what you mean here? do you mean remove the `FeatureFlagSet.of`?
To match everything else which is just `.requiredFeatures(EXPERIMENTAL)`
This cant be done as `MenuType`s ctor specifically wants a `FeatureFlagSet` where as everything else wants `FeatureFlag...` which then is wrapped into a `FeatureFlagSet.of` call

Or do you mean put the comment on the same line as the code?
I put the comment above code here to match the lines above, but i can move the comment if thats desired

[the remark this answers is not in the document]
IchHabeHunger54: Yeah, this is mainly about the comment. Above you always use `.requiredFeatures(EXPERIMENTAL) // comment`, so this should similarly be `FeatureFlagSet.of(EXPERIMENTAL) // comment`

docs/advanced/featureflags.md
```diff
@@ -0,0 +1,280 @@
+# Feature Flags
+
+Feature Flags are a system which allow developers to gate a set of features behind some set of required flags, that being registered elements, gameplay mechanics, data pack entries or some other unique system to your mod.
+
+A common use case would be gating experimental features/elements behind a experimental flag, allowing users to easily switch them on and play around with them before they are finalized.
```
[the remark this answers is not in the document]
ApexModder: what do you mean here exactly?
do you mean all the pale oak blocks are gated cause they depend on the pale oak wood type?
if so thats just simply not true, `WoodType` and its paired `BlockSetType` cannot be gated behind flags, its each individual `Block` thats defining the flag thats gating them, which would fall under `registered elements,`.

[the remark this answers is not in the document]
IchHabeHunger54: No, I just wanted something like "You can also gate your own blocks behind vanilla feature flags. For example if you were to add to the set of Pale Oak wood blocks, you'd only want those to show up if Pale Oak is enabled." (but more general because using Pale Oak as an example is gonna be outdated by the next version)

IchHabeHunger54 (approved):

PR #178, opened 2024-10-14
Update to 1.21.2

ChampionAsh5357:
Applies the contribution formatting and updates the docs to 1.21.2. This is a separate section due to the large number of changes between versions.

### Waiting On
- neoforged/Neoforge#1620
- Xfact's updates to `IClientExtensions` for armor model setup

### Relevant Issues
Closes #177
Closes #179 
Closes #180 
Closes #181
Closes #183
Closes #196 
Supersedes #174

### Subsections

- [x] Advanced
- [x] Block Entities
- [x] Blocks
- [x] Concepts
- [x] Data Storage
- [x] Getting Started
- [x] GUI
- [x] Items
- [x] Legacy
- [x] Misc
- [x] Networking
- [x] Resources
    - [x] Client
        - [x] Models
    - [x] Server
        - [x] Data Maps
        - [x] Enchantments
        - [x] Loot Tables
        - [x] Recipes
- [x] World Generation

------------------
Preview URL: https://pr-178.neoforged-docs-previews.pages.dev

PR #189, opened 2024-11-02
Mention how to target unregistered biomes

TelepathicGrunt:
Was suggested in here to mention how to handle Pale Garden biome
https://github.com/neoforged/NeoForge/issues/1649#issuecomment-2452598540

------------------
Preview URL: https://pr-189.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/worldgen/biomemodifier.md
+++ b/docs/worldgen/biomemodifier.md
@@ -695,7 +695,7 @@ There may be times when a biome modifier needs to target a biome that is not alw
 }
 ```
 
-Using that biome tag for a biome modifier will now not crash if the biome is not registered. A use case for this is the Pale Garden biome in 1.21.3. That biome is only created when you turn on the Winter Drop datapack in-game. Otherwise, the biome does not exist in the biome registry at all. Another use case can be to target modded biomes while still functioning when the other mods not present.
+Using that biome tag for a biome modifier will now not crash if the biome is not registered. A use case for this is the Pale Garden biome in 1.21.3. That biome is only created when you turn on the Winter Drop datapack in-game. Otherwise, the biome does not exist in the biome registry at all. Another use case can be to target modded biomes while still functioning when the mods adding these biomes are not present.
 
 [datareg]: ../concepts/registries.md#datapack-registries
 [staticreg]: ../concepts/registries.md#methods-for-registering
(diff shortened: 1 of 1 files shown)


docs/worldgen/biomemodifier.md
```diff
@@ -695,6 +697,24 @@ This will then result in the following JSON being created:
...
+  "values": [
+    {
+      "id": "minecraft:pale_garden",
+      "required": false
+    }
+  ]
+}
+```
+
+Using that biome tag for a biome modifier will now not crash if the biome is not registered. A use case for this is the Pale Garden biome in 1.21.3. That biome is only created when you turn on the Winter Drop datapack in-game. Otherwise, the biome does not exist in the biome registry at all. Another use case can be to target modded biomes while still functioning when the other mods not present.
```
IchHabeHunger54, line 716:
```suggestion
Using that biome tag for a biome modifier will now not crash if the biome is not registered. A use case for this is the Pale Garden biome in 1.21.3. That biome is only created when you turn on the Winter Drop datapack in-game. Otherwise, the biome does not exist in the biome registry at all. Another use case can be to target modded biomes while still functioning when the mods adding these biomes are not present.
```

IchHabeHunger54 (changes requested):
Looks good, one small nitpick.

IchHabeHunger54 (dismissed):
Approved, but merging only once #178 has been merged as well.

docs/worldgen/biomemodifier.md
```diff
@@ -695,6 +697,24 @@ This will then result in the following JSON being created:
...
+{
+  "replace": false,
+  "values": [
+    {
+      "id": "minecraft:pale_garden",
+      "required": false
+    }
+  ]
+}
+```
```
ChampionAsh5357, line 714: I would also provide a method for how to datagen this using a `TagsProvider`

TelepathicGrunt: I don’t know how to datagen. Would you be able to add a commit to this PR of what that should look like?

ChampionAsh5357:
```java
// In a TagsProvider<Biome> subclass
// Assume we have some example TagKey<Biome> OPTIONAL_BIOMES_TAG
@Override
protected void addTags(HolderLookup.Provider registries) {
    this.tag(OPTIONAL_BIOMES_TAG)
        // Must be a ResourceLocation representing the registry object
        .addOptional(WinterDropBiomes.PALE_GARDEN.location());
}
```

Should be something like that.

IchHabeHunger54:
Rebase, then this should be good to go.

IchHabeHunger54 (approved):

IchHabeHunger54 (approved):

PR #194, opened 2024-11-10
Fixes wrong death message id in DamageTypes docs

starforcraft:
grafik

### DIFF
--- a/docs/resources/server/damagetypes.md
+++ b/docs/resources/server/damagetypes.md
@@ -18,8 +18,8 @@ Now that we can reference it from code, let's specify some properties in the dat
 ```json5
 {
     // The death message id of the damage type. The full death message translation key will be
-    // "death.examplemod.example" (with swapped-out mod ids and names).
-    "message_id": "example",
+    // "death.attack.examplemod.example" (with swapped-out mod id and name).
+    "message_id": "examplemod.example",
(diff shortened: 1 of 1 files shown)


docs/resources/server/damagetypes.md
```diff
@@ -18,7 +18,7 @@ Now that we can reference it from code, let's specify some properties in the dat
 ```json5
 {
   // The death message id of the damage type. The full death message translation key will be
-  // "death.examplemod.example" (with swapped-out mod ids and names).
+  // "death.attack.example" (with swapped-out name).
   "message_id": "example",
```
IchHabeHunger54, line 22:
```suggestion
  // "death.attack.examplemod.example" (with swapped-out mod id and name).
  "message_id": "examplemod.example",
```

IchHabeHunger54 (changes requested):
Thank you for the contribution!

While it is correct that death messages start with `death.attack.` and not just `death.`, the mod id should still be kept somewhere. I've added this as a suggested change.

IchHabeHunger54 (approved):
Approved, will be merged after #178 because we've set a docs freeze until that one is done.

IchHabeHunger54:
Rebase, and then you should be good to go.

IchHabeHunger54 (approved):

PR #197, opened 2024-11-13
Update docs/resources/server/recipes/index.md

Crare:
minor fix. variable "registries" doesn't come anywhere. fixed it by using the incoming method parameter "lookupProvider".


IchHabeHunger54 (approved):
Looks fine, rebase and this should be good to go.

PR #198, opened 2024-11-15
Change feature flags default file location

ApexModder:
Simply changes the default file location for feature flags


IchHabeHunger54 (approved):
For context, we talked about how we should have all these extra JSONs be in META-INF instead of the project root for consistency.

PR #216, opened 2025-01-03
Fix miscapitalized `ByteBufCodecs` in streamcodecs.md

noobanidus:
The class is correctly capitalized `ByteBufCodecs` everywhere else, but in this one example the `Buf` has become `buf`.


docs/networking/streamcodecs.md
sciwhiz12, line 1: Please also fix the same typo at the example under the "Holder Sets" header, some lines down from where you've already made changes. Might as well fix all the incorrect capitalizations in one go.

sciwhiz12 (changes requested):

docs/networking/streamcodecs.md
noobanidus: My bad, I thought I used a case-sensitive search to go through the whole file, but I mustn't have.

sciwhiz12 (approved):

PR #211, opened 2024-12-03
Update to 1.21.4

ChampionAsh5357:
Applies the contribution formatting and updates the docs to 1.21.4.

### Subsections

- [ ] Advanced
- [ ] Block Entities
- [ ] Blocks
- [ ] Concepts
- [ ] Data Storage
- [ ] Getting Started
- [ ] GUI
- [ ] Items
- [ ] Legacy
- [ ] Misc
- [ ] Networking
- [ ] Resources
    - [ ] Client
        - [ ] Models
    - [ ] Server
        - [ ] Data Maps
        - [ ] Enchantments
        - [ ] Loot Tables
        - [ ] Recipes
- [ ] World Generation

PR #217, opened 2025-01-03
Fix small error in Menus doc

starforcraft:


IchHabeHunger54 (approved):
Will be merged after #211 , to avoid having to rebase that PR.

PR #223, opened 2025-01-26
[1.21.4] Update usage of `GatherDataEvent` and the `create*` methods

ChampionAsh5357:
Updates the usage of the `GatherDataEvent` to its current implementation in 1.21.4. Also switches all examples to use one of the available `create*` methods as required.

------------------
Preview URL: https://pr-223.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/resources/index.md
+++ b/docs/resources/index.md
@@ -80,7 +80,13 @@ Data generation, colloquially known as datagen, is a way to programmatically gen
 
 Datagen is run through the Data run configuration, which is generated for you alongside the Client and Server run configurations. The data run configuration follows the [mod lifecycle][lifecycle] until after the registry events are fired. It then fires one of the [`GatherDataEvent`s][event], in which you can register your to-be-generated objects in the form of data providers, writes said objects to disk, and ends the process.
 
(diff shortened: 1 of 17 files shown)


docs/resources/index.md
```diff
@@ -78,7 +78,9 @@ _See also: [`pack.mcmeta` (Resource Pack)][packmcmetaresourcepack] and [`pack.mc
...
+Datagen is run through the Data run configuration, which is generated for you alongside the Client and Server run configurations. The data run configuration follows the [mod lifecycle][lifecycle] until after the registry events are fired. It then fires one of the [`GatherDataEvent`s][event], in which you can register your to-be-generated objects in the form of data providers, writes said objects to disk, and ends the process.
+
+There are two subtypes which operate on the [**physical side**][physicalside]: `GatherDataEvent.Client` and `GatherDataEvent.Server`.  `GatherDataEvent.Client` should contain all providers to generate while `GatherDataEvent.Server` should only contain the providers used to generate datapack entries. As such, all examples shown will use `GatherDataEvent.Client`, as that event is also what is fired by the default `clientData` configuration provided by the MDK.
```
XFactHD, line 83:
```suggestion
There are two subtypes which operate on the [**physical side**][physicalside]: `GatherDataEvent.Client` and `GatherDataEvent.Server`.  `GatherDataEvent.Client` may contain all providers to generate while `GatherDataEvent.Server` may only contain the providers used to generate datapack entries. As such, all examples shown will use `GatherDataEvent.Client`, as that event is also what is fired by the default `clientData` configuration provided by the MDK.
```
If the client datagen event is used for all providers, then there's no point in duplicating the datapack ones in the sever datagen event. The recommendation should be to either put everything in the client one or separate them properly.

XFactHD (approved):

PR #233, opened 2025-03-02
Add support for syntax highlighting of '.properties' files

MelanX:
At the moment, only the [Parchment site](https://docs.neoforged.net/neogradle/docs/parchment) uses a `properties` area, but maybe there will be more in the future.

Before: https://docs.neoforged.net/neogradle/docs/parchment

After: https://pr-233.neoforged-docs-previews.pages.dev/neogradle/docs/parchment

------------------
Preview URL: https://pr-233.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/gettingstarted/modfiles.md
+++ b/docs/gettingstarted/modfiles.md
@@ -45,8 +45,8 @@ While the `group` property in the `build.gradle` is only necessary if you plan t
 
 The group id should be set to your top-level package. See [Packaging][packaging] for more information.
 
-```text
-// In your gradle.properties file
+```properties
+# In your gradle.properties file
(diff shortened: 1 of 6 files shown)


ChampionAsh5357:
Would you mind changing the following line in the docs and its versions to use the `properties` syntax highlighter as well? Best make use of it where it should be used: https://github.com/neoforged/Documentation/blob/604f875ea560417d7b27c31b0d0f0f6b9ee27f21/docs/gettingstarted/modfiles.md?plain=1#L48

MelanX:
Done

ChampionAsh5357 (approved):
Looks good

PR #246, opened 2025-05-21
Give more precisions on how to register a screen with a menu

osechet:
# Description

Reading the documentation about how to register a screen with a menu was a little bit confusing for me, since the `registerScreens` is declared private and non-static.

It seems the usual way to declare this method is in the mod's `ClientModEvents` class as a `public static` method with `@SubscribeEvent`. To show these elements in the documentation can help people to quickly understand how to do.

# Changes made

I updated the "Registering an AbstractContainerScreen" section to make it less confusing.

### DIFF
--- a/docs/gui/screens.md
+++ b/docs/gui/screens.md
@@ -408,11 +408,12 @@ When rendering the label, you do **not** need to specify the `leftPos` and `topP
 
 ## Registering an AbstractContainerScreen
 
-To use an `AbstractContainerScreen` with a menu, it needs to be registered. This can be done by calling `register` within the `RegisterMenuScreensEvent` on the [**mod event bus**][modbus].
+To use an `AbstractContainerScreen` with a menu, it needs to be registered. This can be done by calling `register` within the `RegisterMenuScreensEvent` on the [**mod event bus**][modbus], for example in the `ClientModEvents` class.
 
 ```java
 // Event is listened to on the mod event bus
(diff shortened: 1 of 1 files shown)


osechet:
I just realized it's possible to call the `registerScreens()` method with `modEventBus.addListener(this::registerScreens)` in the mod constructor. Now I don't know what is the official good way to do this.

In any case, I still think the documentation has to be updated either with my proposition or with another showing the `addListener` call.

docs/gui/screens.md
```diff
@@ -408,11 +408,12 @@ When rendering the label, you do **not** need to specify the `leftPos` and `topP
 
 ## Registering an AbstractContainerScreen
 
-To use an `AbstractContainerScreen` with a menu, it needs to be registered. This can be done by calling `register` within the `RegisterMenuScreensEvent` on the [**mod event bus**][modbus].
+To use an `AbstractContainerScreen` with a menu, it needs to be registered. This can be done by calling `register` within the `RegisterMenuScreensEvent` on the [**mod event bus**][modbus], for example in the `ClientModEvents` class.
```
ChampionAsh5357, line 411: You can't really specify a class here since registration differs between users. That's the reason why we link back to the events doc.

ChampionAsh5357 (changes requested):
Thank you for the contribution! Would you mind porting this to the other versions for consistency? Thanks again!

ChampionAsh5357:
Closing as the changes were added in #252

PR #247, opened 2025-05-25
Add "commands" pages (finished)

CCCStudioCoder:
In "data attachments"
Making suppliers private prevents external use, and it's not very partical...
New modders trying these codes might not understand why they can't import the attachments.
In "commands"
Creating and write the complete page
In "modfiles" add a files tree


no-name-provided:
Regarding your submission - I'm not on the team, and I can't speak to their motivations. However, since you asked for feedback, I've taken a look at your pull request and a few things jumped out at me:

1. It appears English isn't your first language - your PR is in need of editing.
2. You have multiple goals - if anyone has a problem with one change, that will slow down acceptance of all changes
3. You initial commit seems trivial to modders who are fluent in Java - as this "isn't a Java wiki", documentation that isn't unique to Minecraft will have low priority.
4. In your change to the modfiles page, your new section begins with a code block. The general trend in this wiki is to begin with a textual introduction, then provide additional information in whatever format helps. Consider rearranging the elements.
5. In the same section, you included incorrect information: "The java folder is always build as java/<com/net>/<username>/<modid> where <com/net> is a [web domains type][https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains] and <username> is your pseudo. Then, in the last child, it's always a <modid>.java file, as ExampleMod.java and a config.java (see [config][config.md])." is wrong.

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
 [versioning]: versioning.md
 [enumextension]: ../advanced/extensibleenums.md
 [featureflags]: ../advanced/featureflags.md
+## Mod files tree
+```
+root
+ |-- .gradle
+ |-- gradle
```
ChampionAsh5357, line 255: Why not expand this folder to show the gradle wrapper?

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
...
+ |-- run
+ |-- src
+ |    |-- main
+ |    |    |-- java
+ |    |    |    |-- <com/net>
+ |    |    |    |    |-- <username>
+ |    |    |    |    |    |-- <modid>
+ |    |    |    |    |    |    |-- <some packages and single files>
+ |    |    |    |    |    |    |-- <modname>.java
+ |    |    |    |    |    |    |-- config.java
```
ChampionAsh5357, line 266: We already have a section on this, but you cannot guarantee that all domains use a `com` or `net` extension. That is why the section specifies what to use.

ChampionAsh5357, line 266: Again, can't assume this file structure, or are you trying to show the example project, in which case this still does not match.

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
...
+ |    |    |-- resources
+ |    |    |    |-- assets
+ |    |    |    |    |-- <modid>
+ |    |    |    |-- data
+ |    |    |    |    |-- minecraft
+ |    |    |    |    |-- <modid>
```
ChampionAsh5357, line 272: Why is there only a `minecraft` for data?

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
...
+ |    |-- test
+ |-- build.gradle
+ |-- gradle.properties
+ |-- gradlew   (file)
+ |-- gradlew.bat
+ |-- settings.gradle
+```
+The `gradle` and `.gradle` folders are Gradle source folders (of the ModDevGradle plugin) that you will not normally need to modify.
```
ChampionAsh5357, line 280: No? The gradle documentation tells you exactly what these folders are for: https://docs.gradle.org/current/userguide/gradle_directories.html#project_root

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
...
+
+The `run` folder contains all data generated by testing your mod (like the `save` folder with all the world data).
```
ChampionAsh5357, line 282: This is poorly explained. You should equate it to the vanilla `.minecraft` folder for a client or where a server generates its files.

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
...
+
+The `src` folder is the center of modding. `test` is optionnal. `java` is the source code and `resources` contains a `àssets` folder, that you've need to build as a [resources pack][https://minecraft.wiki/w/Resource_pack]. `data` is the same as a [datapack][https://minecraft.wiki/w/Data_pack], where <modid> is your mod id (like `examplemod`) and `minecraft` is normally only for build-in tags.
```
ChampionAsh5357, line 284: Using existing links. We have a page for resources and datapacks.

ChampionAsh5357, line 284: Why are we explaining the java and resources folder? It would make more sense to explain Gradle in general since those exist purely because that is how the commonly used build tools format their layouts.

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
...
+
+The java folder is always build as `java/<com/net>/<username>/<modid>` where `<com/net>` is a [web domains type][https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains] and `<username>` is your pseudo. Then, in the last child, it's always a `<modid>.java` file, as `ExampleMod.java` and a `config.java` (see [config][config.md]). Packages are also descripted in [structuring.md].
```
ChampionAsh5357, line 286: This is, once again, an assumption of how the user chooses to structure their files. This is why we have a structuring section.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
+---
+sidebar-position: 7
+---
+# Commands
+Commands are very useful in vanilla Minecraft and even more so with mods. They are the quickest way to access and modify data.
```
ChampionAsh5357, line 5: Commands are meant to perform some operation within the game. Modifying data is only one aspect.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+Commands allow to:
+- Use permission levels.
+- Access to anything in your mod.
+- Allow use anywhere on the map, generally without specific items or other physical objects.
```
ChampionAsh5357, line 9: I don't understand what this means.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+## Creating a command
+To create a command, you need to create a new class and put this code:
```
ChampionAsh5357, line 11: Why? What is it for?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+```java
+public static void register(CommandDispatcher<CommandSourceStack> dispatcher) {
+    dispatcher.register (
+        Commands.literal("commandname") // The name of your command
```
ChampionAsh5357, line 15: Why does it start with a literal?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+        .execute(context -> { // Here is what the command do, where context is a CommandContext<CommandSourceStack>
+            // Put action here
+            return Command.SINGLE_SUCCES; // send succes to server
+        })
```
ChampionAsh5357, line 19: Ok, so the command does something, but what am I taking in, and what am I returning?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+    );
+}
+```
+And then, register the command in your event handler (of anywhere as an event) like this:
```
ChampionAsh5357, line 23: This is on the game event bus.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+```java
+@SubscribeEvent
+public static void onRegisterCommands(RegisterCommandsEvent event) {
```
ChampionAsh5357, line 26: What about the other methods provided by the event?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+    ModCommands.register(event.getDispatcher());
+}
+```
+What Neoforge do in this case, it's that it call event `RegisterCommandEvent` when loading game, and our code say to call `#register` with an `dispatcher` argument, the global source of all commands. Now is the command ready to use!
+## Arguments
+At this point, our command execute always the same action, and it's not very useful...
```
ChampionAsh5357, line 32: This should be written with a documentation mindset foremost. While tutorial aspects are inevitable for understanding, it should not be the highlight of the piece as it ends up lacking in explanation.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+Here is a list of principals argument types:
+- Integer
+- String
+- Entity, BlockPos, Item...
```
ChampionAsh5357, line 36: This seems like a rather arbitrary list. And what do you mean by principal?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+They are located on path `neoforge/net/minecraft/commands/` and native types (int and string) are locked in brigadier class.
```
ChampionAsh5357, line 37: Argument types are in `com.mojang.brigadier.arguments` and `net.minecraft.commands.arguments`, but why does that matter? You should be explaining what the `ArgumentType` interface is and its purpose.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+Argument are invoked by `#then(Commands.argument("name", argument_type))` you can get it with `Type varname = <argument_type>.get<argument>(context, "name")`
+Here is an example:
+```java
+public static void register(CommandDispatcher<CommandSourceStack> dispatcher) {
+    dispatcher.register (
+        Commands.literal("commandname")
+        .then(Commands.argument("message", StringArgumentType.string())) // or word(), it's the string type
```
ChampionAsh5357, line 44: String and word are different. What are you specifically trying to get?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+```
+What Neoforge do in this case, it's that it call event `RegisterCommandEvent` when loading game, and our code say to call `#register` with an `dispatcher` argument, the global source of all commands. Now is the command ready to use!
+## Arguments
+At this point, our command execute always the same action, and it's not very useful...
+Here is a list of principals argument types:
+- Integer
+- String
+- Entity, BlockPos, Item...
+They are located on path `neoforge/net/minecraft/commands/` and native types (int and string) are locked in brigadier class.
+Argument are invoked by `#then(Commands.argument("name", argument_type))` you can get it with `Type varname = <argument_type>.get<argument>(context, "name")`
```
ChampionAsh5357, line 38: This is not true for all argument types.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+Here is an example:
+```java
+public static void register(CommandDispatcher<CommandSourceStack> dispatcher) {
+    dispatcher.register (
+        Commands.literal("commandname")
+        .then(Commands.argument("message", StringArgumentType.string())) // or word(), it's the string type
+        .execute(context -> {
+            String message = StringArgumentType.getString(context, "message"); // Same name as above
+            ServerPlayer player = context.getSource(); // Who is executing the command
```
ChampionAsh5357, line 47: You mentioned above that the source is the `CommandSourceStack`. Why is it now the player?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+            player.sendServerMessage(message);
```
ChampionAsh5357, line 48: `ServerPlayer#sendServerMessage` is not a method. Are you thinking of `sendSystemMessage`? Even then, it doesn't take in a string but a `Component`.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+            return Command.SINGLE_SUCCES;
+        })
+    );
+}
+```
+:::danger
+Note that all ArgumentTypes are abstract and you need to use child, as `#string()` in this example.
```
ChampionAsh5357, line 55: What does this mean??? Argument types are not abstract.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+:::
+### Suggestion
+You can suggest (implement auto-completion) things with method `#suggest(SuggestionProvider)` just after a `#then` expression.
```
ChampionAsh5357, line 58: No, the `suggests` (notice the `s`) method should be called following some argument if you want to override the argument's suggestions provider. `then` is for something completely different.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+    SharedSuggestionProvider.suggest(ACTIONS, builder); // We use ACTIONS list
+```
+:::note
+Users aren't obliged to use suggestions. If you want to error when gived argument doesn't match with suggestion, you can do in the execute statement:
+```java
+if(!ACTIONS.contains(arg)) { // Arg is already declared
+    context.getSource().sendFailure(Component.literal("Invalid action (add/sub/set)")); // As return send_succes method, but it send failure (red message)
+    return 0; // Ensure to stop the function
+}
+```
```
ChampionAsh5357, line 74: I don't understand the purpose of this example as brigadier will already create suggestions for subcommands.

ChampionAsh5357, line 74: Why would you do this? Brigadier already tells you if it is an unknown command or if you don't have the permissions to run it.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+:::
+## Other features
+As many Neoforge elements, command has his own set of function that make our life better...
+### Execution conditions
+You can use the `#requires` method after `Commands.literal` function to require some things with based on an function that take one argument: `context`.
```
ChampionAsh5357, line 79: This and all others similar should be reframed in the context of the `ArgumentBuilder`, which is how the commands are built.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+```java
+// In the register function
+dispatcher.register(
+    Commands.literal("opcommand")
+        .requires(context -> context.hasPermission(2))
```
ChampionAsh5357, line 84: What does the 2 mean? Don't simply use things without explaining them.

docs/misc/commands.md
ChampionAsh5357, line 79: It's a predicate that takes in the command source and returns whether the source meets the requirements to run the command.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+);
+```
+:::note
+Of course, you can place `#requires` in a `if` statement or only when certains parameters equals to somethings.
```
ChampionAsh5357, line 88: What does this mean?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+:::
+### Side requirement
+In your `register` method, you have a parameter `dispatcher`. In addition to enabling registration, it gave some informations about tick game context. One of these information is `#getCommandSelection`, which returns whether it is multiplayer or singleplayer.
```
ChampionAsh5357, line 91: This is what the event provides, not your static method.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+```java
+public static void register(CommandDispatcher<CommandSourceStack> dispatcher) {
+        if(event.getCommandSelection() == Commands.CommandSelection.DEDICATED) { // INTEGRATED for singleplayer
```
ChampionAsh5357, line 94: `event` is not a valid field.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+            dispatcher.register(
+                Commands.literal("multiplayercommand")
+                        // then code action
+            );
+        }
+    }
+```
```
ChampionAsh5357, line 101: What happened to the command build context?

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+### `Commands.literal`
```
ChampionAsh5357, line 102: Are you trying to explain the subcommand structure? What does this have to do with `Commands#literal`?  This goes back to the `ArgumentBuilder`.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+        .then(Commands.literal("get"
+            .executes(ctx -> {}) // do somethings
+        )
+        .then(Commands.argument("action"), StringArgumentType.word())
+            .then(Commands.argument("value"), IntegerArgumentType.integer())
+                .executes(ctx -> {}) // do also somethings
+        // then maybe other code actions
+    );
+```
+Here, if the first argument is `get`, we directy execute the command, otherwise we take another argument `value`.
```
ChampionAsh5357, line 116: No, this provides the following commands: `/modifyorget get` or `/modifyorget <action> <value>`.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+:::danger
+If you use `Commands.literal`, you will always need to put related instructions in this `then` block. Otherwise will you get an error of a strange behavior.
```
ChampionAsh5357, line 118: I have no idea what this is referring to.

docs/misc/commands.md
```diff
@@ -0,0 +1,122 @@
...
+:::
+:::note
+Now you have all stuff needed to create your own command. Remind that all structures are possible, and you can make evevrythings like invoking suggestion only if...
```
ChampionAsh5357, line 121: What does the reminder mean?

CCCStudioCoder:
Thank you for all your reply! I will try to correct all of that.

> 1. It appears English isn't your first language - your PR is in need of editing.
> 2. You have multiple goals - if anyone has a problem with one change, that will slow down acceptance of all changes
> 3. You initial commit seems trivial to modders who are fluent in Java - as this "isn't a Java wiki", documentation that isn't unique to Minecraft will have low priority.
> 4. In your change to the modfiles page, your new section begins with a code block. The general trend in this wiki is to begin with a textual introduction, then provide additional information in whatever format helps. Consider rearranging the elements.
> 5. In the same section, you included incorrect information: "The java folder is always build as java/<com/net>// where <com/net> is a [web domains type][https://en.wikipedia.org/wiki/List_of_Internet_top-level_domains] and  is your pseudo. Then, in the last child, it's always a .java file, as ExampleMod.java and a config.java (see [config][config.md])." is wrong.

1: Yes, it isn't my first langage, i do what i can.
2: I don't understand where you're going.
3: It is still true, and it could mislead new modders. Although it is not a priority, there is no downside to adding it.
4: I'm correcting this.
5: Idem.

docs/misc/commands.md
CCCStudioCoder: Completely forgot it, sorry!

docs/misc/commands.md
CCCStudioCoder: That's true, but it's another way of explaining it. You're right, I'll change it anyway.

docs/misc/commands.md
CCCStudioCoder: If you use commands.literal as single-possibility argument, [...

docs/misc/commands.md
CCCStudioCoder: doesn't exist anymore...

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
 [versioning]: versioning.md
 [enumextension]: ../advanced/extensibleenums.md
 [featureflags]: ../advanced/featureflags.md
+## Mod files tree
+```
+root
```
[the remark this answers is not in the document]
CCCStudioCoder: changed! It's the mod root folder

docs/gettingstarted/modfiles.md
CCCStudioCoder: all this folders are now removed from this file tree!

docs/misc/commands.md
CCCStudioCoder: it was an example

docs/misc/commands.md
CCCStudioCoder: Once again, I went too fast.

docs/misc/commands.md
CCCStudioCoder: pas tous, les selector sont en effet automatiquement suggerés, mais pas les strings

CCCStudioCoder: for strings, it does not detect if it is not a good argument for strings

docs/gettingstarted/modfiles.md
```diff
@@ -248,3 +248,39 @@ An entry in `neoforge.mods.toml` does not need a corresponding `@Mod` annotation
 [versioning]: versioning.md
 [enumextension]: ../advanced/extensibleenums.md
 [featureflags]: ../advanced/featureflags.md
+## Mod files tree
```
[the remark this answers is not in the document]
ChampionAsh5357: Look at the line literally above the title. Those are the references that should be at the bottom of the page.

docs/misc/commands.md
ChampionAsh5357: Sure, but the event should be explained either before everything or after everything. Before probably makes the most sense as the parameters you need to register a command are provided by the event.

docs/misc/commands.md
ChampionAsh5357: That's the same with all brigadier argument types (except bool). The way the example reads makes me feel like you're trying to do `/<some command> add <do something>`, which, unless you have some sort of dynamic command structure, is not the common usecase.

ChampionAsh5357: This goes back to me thinking your example is trying to do something with subcommands than strings. Additionally, your commands should be throwing exceptions if there's something like an invalid action instead of handling the failure result yourself.

docs/misc/commands.md
CCCStudioCoder: changed

docs/misc/commands.md
CCCStudioCoder: For "/execute" vanilla command. It's string type for statements like run, as, at... Without `#suggest', there's no suggestion when typing the command.

docs/misc/commands.md
ChampionAsh5357: But that's not explained. I could easily pass in an argument, but why should I start with a `literal`. It is not necessary for `register`, but it is the best practice to do so.

docs/misc/commands.md
ChampionAsh5357: No, I mean that the event provides the command selection and build context which you could be using in your static method, so the event should either be explained first or the sections should be expanded upon when they are used.

docs/misc/commands.md
ChampionAsh5357: What? No it isn't. Those are each separate subcommands. You can check the `ExecuteCommand` yourself. The only suggestions that are overridden are for a subset of available entity types, score holders, boss events, loot item predicates, and the mcfunctions.

CCCStudioCoder:
can somebody clarify what need i to change now?

docs/misc/commands.md
ChampionAsh5357: You still haven't really explained what a 'success' is. Why is success 1? Can it be larger numbers? Smaller numbers? You need to explain.

docs/misc/commands.md
ChampionAsh5357: Once again, they both take in different arguments for a string when parsing. You cannot just brush it off as the command itself can look different. For example, both word and string can take in `hello`, but only string can take in `"hello world"`

docs/misc/commands.md
ChampionAsh5357: You can't resolve messages if you do not address the issue.

docs/misc/commands.md
ChampionAsh5357: The build context is still not explained.

docs/misc/commands.md
ChampionAsh5357: It's not another way of explaining it. `/modifyorget get <value>` is not a valid command under this structure.

PR #269, opened 2025-08-25
Creating "Commands" page in docs/misc

CCCStudioCoder:
A clear and complete command page :)

### DIFF
--- a/docs/misc/commands.md
+++ b/docs/misc/commands.md
@@ -1,0 +1,128 @@
+# Commands
+Commands are very useful because it provides a way for players to test
+your mod faster and cheat when he doesn't want to spend a long time to do
+not instresting things.
+
+Mojang provides a complete command system in the `com.mojang.brigadier` package, which includes
+ - Command parsing and handling
+ - Primitive [argument types](#argument-types)
+ - [Command Contexts](#command-contexts)
+ - Suggestions
+ - [Exception handling](#exceptions)
+
+## Commands architecture
+All the pieces of commands creation.
+:::tip
+See [Creating a command](#creating-a-command) for intuitive examples.
+:::
+
+### Command Dispatcher
+The `CommandDispatcher<S>` (`S` is `CommandSourceStack` in this case) is like a [`DeferredRegister`](../concepts/registries.md), but it's a singleton.
+Its function `register` work with a `LiteralArgumentBuilder`.
+
+### Argument Builder
+The `ArgumentBuilder` class is abstract, so it can't be used on its own.
+It has two children:
+ - `LiteralArgumentBuilder`: the most used of the two, because `CommandDispatcher#register` only accepts it. Returned after literal arguments and executions
+ - RequiredArgumentBuilder: used in rare cases. It has more parameters and associated get functions than `LiteralArgumentBuilder`. Returned after required arguments declarations.
+
+In fact, the entire commands are built with it. Almost every function in the `Commands` class returns an `ArgumentBuilder`.
+
+Last but not least, command trees are chains of builders created with some functions:
+ - `#then(ArgumentBuilder<S, ?>)`: create a new branch.
+ - `#executes(Function<CommandContext<CommandSourceStack>, Integer>)`: execute the command. You'll need to return an integer: `Command.SINGLE_SUCCESS` (or just `1`) for success and `0` otherwise.
+
+### Argument types
+Argument types is the second part of creating commands. They come in two packages:
+ - `com.mojang.brigadier.arguments` for the argument types based on [Primitive Types](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html#PageTitle) (including `String`). Their names are always `<type>ArgumentType`.
+ - `net.minecraft.commands.arguments` for all other argument types. Their names are always `<type>Argument`.
+
+All these argument types are implementing an interface: `ArgumentType<T>` where `T` is the type on which they are based.
+
+When you create an argument for your commands, you will always need to choose an `ArgumentType`.
+
+### Command Contexts
+Command contexts are used every time you run a command. It's based on a generic type, the additional information, often `CommandSourceStack`.
+It contains all the additional information plus what was written as arguments.
+
+### Exceptions
+The package `com.mojang.brigadier.exceptions` provide one [exception](https://docs.oracle.com/javase/tutorial/essential/exceptions/index.html): `CommandSyntaxException`, throwed if the command isn't correctly written; and many internal handling classes.
+:::note
+Minecraft also provides a `FunctionInstanciationsException`, but command functions aren't very useful and clear for external utilisations.
+:::
+
+### Command Source Stack
+The `CommandSourceStack' holds all the information about the command execution context **before** the command is written and executed.
+
+The most important ones are
+ - `boolean #hasPermission(int)`: returns whether the command executor has the given [permission level](https://minecraft.wiki/w/Permission_level).
+ - `Entity #getEntity()`: returns who is executing the command.
+ - `void #sendSuccess(Supplier<Component>, boolean)`: send a green success message to the executor. The supplier provides the message and the boolean if it's visible to everyone.
+ - `void #sendFailure(Component)`: send a red failure message to the executor.
+
+### The `Commands` class
+It provides many tools like `Commands.argument`.
+
+## Creating a command
+To create a command, you'll need to register it with `RegisterCommandsEvent`.
+```java
+@SubscribeEvent
+public static void registerCommands(RegisterCommandsEvent event) {
+    event.getDispatcher().register( // event.getDispatcher returns the CommandDispatcher
+        Commands.literal("mycommand") // Always start with this, the name of your command
+            .requires(source -> source.hasPermission(2)) // Check permissions level
+            .then(Commands.argument("level", StringArgumentType.word())) // Here we create an on-word argument named "level"
+            .executes(context -> {
+                context.getPlayer().sendServerMessage(Component.literal("You're now on the level: " + 
+                StringArgumentType.getString(context, "level")))); // And we send a message to the executor
+                return Command.SINGLE_SUCCESS;
+            }
+    )
+}
+```
+:::warning
+**Do not use special characters or spaces as arguments** in `Commands.literal` and `Commands.argument`. It won't make the game crash, but it will make the command unusable and always give errors.
+:::
+:::tip
+You can create optional argument like this:
+```java
+dispatcher.register( // Assuming we already have a variable holding this dispatcher
+    Commands.literal("mycommandwithoptionalargs")
+    .then(Commands.argument("level", StringArgumentType.word())
+    .executes(ctx -> {/*execution with one argument*/}) // Here #executes is in the "then" block
+    )
+    .then(Commands.argument("message", StringArgumentType.string())
+    .executes(ctx -> {/*execution with two arguments*/})
+)
+```
+:::
+## Creating argument types
+If you wanna add custom argument type like for quests, spell or anythings new that your mod add read below.
+
+To create an argument type, you'll need to create a new class that implements `ArgumentType<T>` and its functions.
+```java
+public class CarArgumentType implements ArgumentType<Car> { // We create argument type based on the class Car
+    public Car parse(StringReader reader) { // To parse the written string to a car instance
+        return CARS.get(reader.read()); // If we have a HashMap<String, Car>
+    }
+    public <S> CompletableFuture<Suggestion> listSuggestion(CommandContext<S>, SuggestionBuilder builder) { // To suggest all the different cars
+        return SharedSuggestionProvider.suggest(CARS.keySet(), builder)
+    }
+    public Collection<String> getExamples() {
+        return List.of("Volvo", "Citroën");
+    }
+}
+```
+
+You need also to add some functions:
+ - A `public static CarArgumentType car()` that returns a `new CarArgumentType()` (it's a best pratice to avoid direct instanciations).
+ - A function to get our arguments:
+ ```java
+ public static Car getCar(CommandContext<CommandSourceStack> ctx, String name) {
+    return ctx.getArgument(name, Car.class);
+ }
+ ```
+
+:::tip
+If you just wanna have custom suggestions like for an action (add, set, sub...) you can directly use `#suggest` after a `then` with as argument an equivalent of `CarAgumentType::listSuggestion`
+:::


docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+public class CarArgumentType implements ArgumentType<Car> { // We create argument type based on the class Car
+    public Car parse(StringReader reader) { // To parse the written string to a car instance
+        return CARS.get(reader.read()); // If we have a HashMap<String, Car>
+    }
+    public <S> CompletableFuture<Suggestion> listSuggestion(CommandContext<S>, SuggestionBuilder builder) { // To suggest all the different cars
+        return SharedSuggestionProvider.suggest(CARS.keySet(), builder)
+    }
+    public Collection<String> getExamples() {
+        return List.of("Volvo", "Citroën");
+    }
```
ChampionAsh5357, line 113: What are the examples for? How do they different from the suggestions?

ChampionAsh5357 (changes requested):
This is a reopen of #247.

Like the previous iteration, this does not really explain what the brigadier system and how its used, more reading like 'these methods are available, and here is an example'. There needs to be a lot more explanation and descriptions about every step in the command structure.

Additionally, the document needs another pass for spelling and grammar. I did not mention them in the review, but there were quite a number.

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+ - A `public static CarArgumentType car()` that returns a `new CarArgumentType()` (it's a best pratice to avoid direct instanciations).
+ - A function to get our arguments:
+ ```java
+ public static Car getCar(CommandContext<CommandSourceStack> ctx, String name) {
+    return ctx.getArgument(name, Car.class);
+ }
+ ```
+
+:::tip
+If you just wanna have custom suggestions like for an action (add, set, sub...) you can directly use `#suggest` after a `then` with as argument an equivalent of `CarAgumentType::listSuggestion`
```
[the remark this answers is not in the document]
CCCStudioCoder: No, it doesn't, you can use `#suggest(SuggestionProvider)`

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+The `ArgumentBuilder` class is abstract, so it can't be used on its own.
+It has two children:
+ - `LiteralArgumentBuilder`: the most used of the two, because `CommandDispatcher#register` only accepts it. Returned after literal arguments and executions
+ - RequiredArgumentBuilder: used in rare cases. It has more parameters and associated get functions than `LiteralArgumentBuilder`. Returned after required arguments declarations.
+
+In fact, the entire commands are built with it. Almost every function in the `Commands` class returns an `ArgumentBuilder`.
+
+Last but not least, command trees are chains of builders created with some functions:
+ - `#then(ArgumentBuilder<S, ?>)`: create a new branch.
+ - `#executes(Function<CommandContext<CommandSourceStack>, Integer>)`: execute the command. You'll need to return an integer: `Command.SINGLE_SUCCESS` (or just `1`) for success and `0` otherwise.
```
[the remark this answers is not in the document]
CCCStudioCoder: sorry, it was because i write Command often like Function

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Argument types
+Argument types is the second part of creating commands. They come in two packages:
```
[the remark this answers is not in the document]
CCCStudioCoder: are you sure you point the good paragraph? I don't see what are you meaning

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+ - `com.mojang.brigadier.arguments` for the argument types based on [Primitive Types](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html#PageTitle) (including `String`). Their names are always `<type>ArgumentType`.
+ - `net.minecraft.commands.arguments` for all other argument types. Their names are always `<type>Argument`.
```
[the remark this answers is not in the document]
CCCStudioCoder: i just learn its existence...

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+All these argument types are implementing an interface: `ArgumentType<T>` where `T` is the type on which they are based.
+
+When you create an argument for your commands, you will always need to choose an `ArgumentType`.
+
+### Command Contexts
+Command contexts are used every time you run a command. It's based on a generic type, the additional information, often `CommandSourceStack`.
+It contains all the additional information plus what was written as arguments.
+
+### Exceptions
+The package `com.mojang.brigadier.exceptions` provide one [exception](https://docs.oracle.com/javase/tutorial/essential/exceptions/index.html): `CommandSyntaxException`, throwed if the command isn't correctly written; and many internal handling classes.
```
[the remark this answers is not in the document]
CCCStudioCoder: What would you write in my place?

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+```java
+@SubscribeEvent
+public static void registerCommands(RegisterCommandsEvent event) {
+    event.getDispatcher().register( // event.getDispatcher returns the CommandDispatcher
+        Commands.literal("mycommand") // Always start with this, the name of your command
+            .requires(source -> source.hasPermission(2)) // Check permissions level
+            .then(Commands.argument("level", StringArgumentType.word())) // Here we create an on-word argument named "level"
+            .executes(context -> {
+                context.getPlayer().sendServerMessage(Component.literal("You're now on the level: " + 
+                StringArgumentType.getString(context, "level")))); // And we send a message to the executor
```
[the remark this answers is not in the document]
CCCStudioCoder: nobody will copy paste this code or check for errors

CCCStudioCoder:
> This is a reopen of #247.
> 
> Like the previous iteration, this does not really explain what the brigadier system and how its used, more reading like 'these methods are available, and here is an example'. There needs to be a lot more explanation and descriptions about every step in the command structure.
> 
> Additionally, the document needs another pass for spelling and grammar. I did not mention them in the review, but there were quite a number.

help me to find grammar mistakes

docs/misc/commands.md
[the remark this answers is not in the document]
ChampionAsh5357: The way you say second part of creating commands indicates to me that you can't use literals after the root node. That is my understanding reading the sentence.

docs/misc/commands.md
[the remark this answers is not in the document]
ChampionAsh5357: 'Whenever a commands errors out, whether from a malformed input or a execution with no successful operation, a `CommandSyntaxException` is thrown. `CommandSyntaxException` is a special kind of exception as it doesn't crash the game, instead sending the error message to the source's out box (e.g., the chat box for players or the output of a command block).'

Not those exact words, but something similar.

docs/misc/commands.md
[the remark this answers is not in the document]
ChampionAsh5357: You'd be surprised how many people do exactly that. Plus, code examples should never be inaccurate if possible. Things may be omitted, but never inaccurate.

docs/misc/commands.md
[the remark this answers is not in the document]
ChampionAsh5357: First, it's `suggests`, and second, yes, that method only exists if the current node is a `RequiredArgumentBuilder`, which is an argument. `suggests` does not exist for literal nodes.

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+                return Command.SINGLE_SUCCESS;
+            }
+    )
```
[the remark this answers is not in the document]
CCCStudioCoder: ohhhhhhhh, yes. I search it for an hour...

docs/misc/commands.md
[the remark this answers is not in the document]
CCCStudioCoder: So I can just specify it.

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+:::note
+Minecraft also provides a `FunctionInstanciationsException`, but command functions aren't very useful and clear for external utilisations.
+:::
+
+### Command Source Stack
+The `CommandSourceStack' holds all the information about the command execution context **before** the command is written and executed.
+
+The most important ones are
+ - `boolean #hasPermission(int)`: returns whether the command executor has the given [permission level](https://minecraft.wiki/w/Permission_level).
+ - `Entity #getEntity()`: returns who is executing the command.
```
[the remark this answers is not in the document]
CCCStudioCoder: and?

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+    .then(Commands.argument("level", StringArgumentType.word())
+    .executes(ctx -> {/*execution with one argument*/}) // Here #executes is in the "then" block
+    )
+    .then(Commands.argument("message", StringArgumentType.string())
+    .executes(ctx -> {/*execution with two arguments*/})
+)
+```
+:::
+## Creating argument types
+If you wanna add custom argument type like for quests, spell or anythings new that your mod add read below.
```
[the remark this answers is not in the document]
ChampionAsh5357: It's important, but you need to use more nuanced terms. For example most registry objects don't need their own argument type, and can instead use the resource key argument. You need to specifically mention to check existing arguments to avoid duplication, or provide some examples.

docs/misc/commands.md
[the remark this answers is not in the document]
ChampionAsh5357: You're considering it to be important, but I believe that's less important than correctly operating on the desired source.

CCCStudioCoder:
@ChampionAsh5357 I just pushed review 3. What else would you say?

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+ - [Exception handling](#exceptions)
+ - Some other small things.
+
+ In fact, Brigadier is the command API of Minecraft.
+
+## Commands architecture
+Here are the main classes and concepts that command creation has.
+
+### Command Dispatcher
+The `CommandDispatcher<S>` (`S` is often `CommandSourceStack` in this case) work like a [`DeferredRegister`](../concepts/registries.md), but it's internally created.
```
ChampionAsh5357, line 23: It still does not work like a `DeferredRegister`. You are not providing a registry name or anything, you are just building the command path.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+Its function `register` require a `LiteralArgumentBuilder` argument.
+
+### Argument Builder
+The `ArgumentBuilder` class is abstract, so it can't be used on its own.
+It has two children:
+ - `LiteralArgumentBuilder`: `CommandDispatcher#register` only accepts it. Returned after literal arguments (one-possiblity) and executions
```
ChampionAsh5357, line 29: This does not explain what a literal argument is. It can be used anywhere as part of the command, though it is typically as the root a the initial sublevels.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+ - RequiredArgumentBuilder: it has more parameters and associated get functions than `LiteralArgumentBuilder`. Returned after required arguments declarations.
```
ChampionAsh5357, line 30: ???? Why is that the most important thing? Required arguments are meant to provide a type of input to be accepted as the next entry.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+In fact, the entire commands are built with it.
+Last but not least, command trees are chains of builders created with some functions:
```
ChampionAsh5357, line 33: There are a lot more than the functions mentioned below.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+ - `#then(ArgumentBuilder<S, ?>)`: create a new argument.
+ - `#executes(Command)`: execute the command. `Command` can be written like a `Function<CommandContext<CommandSourceStack>, Integer>` where you need to return an integer: `Command.SINGLE_SUCCESS` (or just `1`) for success and `0` otherwise.
```
ChampionAsh5357, line 35: This is still wrong. The return value represents how many time the command performed the desired operation.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Argument types
+Argument types is the second part of creating commands. They come in two packages:
+ - `com.mojang.brigadier.arguments` for the argument types based on [Primitive Types](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html#PageTitle) (including `String`, without `char`). Their names are always `<type>ArgumentType`.
+ - `net.minecraft.commands.arguments` for all other argument types. Their names are always `<type>Argument`.
+ - `net.neoforged.neoforge.server.command`: for mod-oriented argument like `EnumArgument`. Their names are the same as for `net.minecraft`.
```
ChampionAsh5357, line 41: They are not for mod-oriented arguments, just say NeoForge provides additional command arguments. Also, the last sentence should be rephrased better, mentioning the naming schema rather than the name itself.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+All these argument types are implementing an interface: `ArgumentType<T>` where `T` is the type on which they are based.
+
+When you create an argument for your commands, you will always need to choose an `ArgumentType`.
```
ChampionAsh5357, line 45: Probably better to mention the required argument builder specifically.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+Command contexts are used every time you run a command. It's based on a generic type, the additional information, often `CommandSourceStack`.
+It contains all the additional information plus what was written as arguments.
+
+### Exceptions
+The package `com.mojang.brigadier.exceptions` provide one [exception](https://docs.oracle.com/javase/tutorial/essential/exceptions/index.html): `CommandSyntaxException`, will be thrown if the syntax is incorrect or if the command fails. It will give an error message in chat if it was a player executing the command, or as output in a [command block](https://minecraft.wiki/w/command_block) if it's a command block.
+
+It also provides many internal handler classes and the `CommandExceptionType`s.
+
+#### `CommandExceptionType`
+This is an empty interface is used to mark a class as a eponymous exception. `CommandSyntaxException` take an argument `CommandExceptionType`.
```
ChampionAsh5357, line 57: This doesn't explain what the exception type is at all. The reason why the exception type is passed in is purely to check against specific exception types, such as if you don't want the command to fail and fallback when a specific exception is thrown.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+The classes implementing `CommandExceptionType` aims to create exception based on a certain amount of parameters:*
```
ChampionAsh5357, line 59: The exception types provide a way to take the erroring objects as an input to the failure message.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+ - `DynamicCommandExceptionType`: one
+ - `Dynamic2CommandExceptionType`, `Dynamic3CommandExceptionType`, `Dynamic4CommandExceptionType`: 2, 3 or 4
+ - `DynamicNCommandExceptionType`: an undetermined amount
+ - `SimpleCommandExceptionType`: zero
+
+These classes are most used in `BuiltInExceptions` in lines like:
```
ChampionAsh5357, line 65: No? There are plenty in the command classes. This is just common exceptions, sort of like how `Codec` contains common codecs rather than all codecs.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+```java
+private static final Dynamic2CommandExceptionType FLOAT_TOO_BIG = new Dynamic2CommandExceptionType((found, max) -> new LiteralMessage("Float must not be
+    more than " + max + ", found " + found));
+```
+and then
+```java
+public Dynamic2CommandExceptionType floatTooHigh() {
+    return FLOAT_TOO_BIG;
+}
+```
```
ChampionAsh5357, line 75: Wait why? This is completely unnecessary.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+:::tip
+These functions are accessible everywhere, and it's recommended to use it when throwing `CommandSyntaxException`.
+```java
+throw new CommandSyntaxException(BuiltInException.floatTooHigh(), new LiteralMessage("Please provide a smaller float!"));
+```
```
ChampionAsh5357, line 81: This is not how you use the exception types. You construct the exception using `create` or `createWithContext`

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+### `SharedSuggestionProvider`
+It's the base interface of execution contexts, [client-side or server-side](../concepts/sides.md).
+
+:::warning
+We talk here about **execution** context. It's meaning that the command hasn't be written when it's instanciated, so you can't get arguments from it, use instead [`CommandContext`](#command-contexts).
+:::
+
+#### `ClientSuggestionProvider`
+It implement `SharedSuggestionProvider` but have many more function, about client context.
```
ChampionAsh5357, line 92: This does not explain what the client suggestion provider is. I don't even think it needs its own section, as it should be combined with either the client commands docs or how the command system works.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+throw new CommandSyntaxException(BuiltInException.floatTooHigh(), new LiteralMessage("Please provide a smaller float!"));
+```
+:::
+
+### `SharedSuggestionProvider`
+It's the base interface of execution contexts, [client-side or server-side](../concepts/sides.md).
+
+:::warning
+We talk here about **execution** context. It's meaning that the command hasn't be written when it's instanciated, so you can't get arguments from it, use instead [`CommandContext`](#command-contexts).
+:::
```
ChampionAsh5357, line 89: Yes but also no. It does happen to be what both use, but I wouldn't consider it the execution context. `ExecutionCommandSource` is what I would say that's for. The suggestion provider is typically implemented on the command sources as that is what provides the helpers to properly handle suggestions for an argument.

docs/misc/commands.md
```diff
@@ -0,0 +1,129 @@
...
+The package `com.mojang.brigadier.exceptions` provide one [exception](https://docs.oracle.com/javase/tutorial/essential/exceptions/index.html): `CommandSyntaxException`, throwed if the command isn't correctly written; and many internal handling classes.
+:::note
+Minecraft also provides a `FunctionInstanciationsException`, but command functions aren't very useful and clear for external utilisations.
+:::
+
+### Command Source Stack
+The `CommandSourceStack' holds all the information about the command execution context **before** the command is written and executed.
+
+The most important ones are
+ - `boolean #hasPermission(int)`: returns whether the command executor has the given [permission level](https://minecraft.wiki/w/Permission_level).
```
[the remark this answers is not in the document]
ChampionAsh5357: What about NeoForge's permission API? It doesn't matter about practicality, but you need to be accurate when writing docs. Also, while external resources are useful, the point of the docs is to have that information in one place. We may end up duplicating stuff from other resources, but we add more context for NeoForge.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+ - `void #sendSuccess(Supplier<Component>, boolean)`: send a success message to the executor. The supplier provides the message and the boolean if it's visible to everyone.
+ - `void #sendFailure(Component)`: send a red failure message to the executor.
+
+#### `ClientCommandSourceStack`
+It's a Neoforge class that implements `SharedSuggestionProvider`. It has no other functions, so it works like the base context.
```
ChampionAsh5357, line 103: The `ClientCommandSourceStack` is meant as as a way to handle commands executed on the client rather than passed from the server.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+### The `Commands` class
+It's the "util" class of Minecraft's commands. You can use some of its function for yours, like `Commands.argument`.
+It also work as the register of all Minecraft's and Neoforge's commands, but you can't use it without [Mixin](https://spongepowered.org) (not recommended).
```
ChampionAsh5357, line 107: What is this supposed to mean?

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Creating a command
+To create a command, you'll need to register it with `RegisterCommandsEvent` [event](../concepts/events.md). Here's an example:
+```java
+// Assuming we already regsiter the event
+@SubscribeEvent // Default event bus
+public static void registerCommands(RegisterCommandsEvent event) {
```
ChampionAsh5357, line 114: Please see #252 for EBS standardization

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+    event.getDispatcher().register( // event.getDispatcher returns the CommandDispatcher
+        Commands.literal("mycommand") // Always start with this, the name of your command
+            .requires(source -> source.hasPermission(2)) // Check permissions level
+            .then(Commands.argument("level", StringArgumentType.word())) // Here we create an on-word argument named "level"
+            .executes(context -> {
+                if(context.getEntity() instanceof Player player) {
+                    player.sendServerMessage(Component.literal("You're now on the level: " + 
+                        StringArgumentType.getString(context, "level"))); // And we send a message to the executor
+                }
+                return Command.SINGLE_SUCCESS;
```
ChampionAsh5357, line 124: Why success if the command doesn't do anything if its not from a player? It would be more useful to return an error message.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+            })
+    );
+}
+```
+:::warning
+**Do not use special characters or spaces** in `Commands.argument` as the name of the argument or in `Commands.literal` as the name of the command. It won't make the game crash, but it will make the command unusable and always give errors.
```
ChampionAsh5357, line 130: Again, special characters can be used. Why do you think it can't be?

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+:::
+:::tip
+You can create optional argument like this:
+```java
+dispatcher.register( // Assuming we already have a variable holding this dispatcher
+    Commands.literal("mycommandwithoptionalargs")
+    .then(Commands.argument("level", StringArgumentType.word())
+    .then(Commands.argument("amount", IntegerArgumentType.integer()))
+    .executes(/*execution with two arguments*/)) // The executes block is inside the then block
+    .executes(/*execution with one argument*/) // The executes block is after the then block
```
ChampionAsh5357, line 140: Use tabs to more easily differentiate the level.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+);
+```
+:::
+## Creating argument types
+Sometime creating a `StringArgumentType` or `ResourceOrTagKeyArgument` isn't pratice or useful.
```
ChampionAsh5357, line 145: Wait why? You can't just say things without explaining.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+In this case you can create your own.
+
+To create an argument type, you'll need to create a new class that implements `ArgumentType<T>` and its functions.
+```java
+public class CarArgumentType implements ArgumentType<Car> { // We create argument type based on the class Car
+    public Car parse(StringReader reader) { // To parse the written string to a car instance
+        return CARS.get(ResourceLocation.read(reader)); // If we have a HashMap<ResourceLocation, Car>
+    }
+    public <S> CompletableFuture<Suggestion> listSuggestion(CommandContext<S> ctx, SuggestionBuilder builder) { // To suggest all the different cars
+        return SharedSuggestionProvider.suggest(CARS.keySet().map(key::toString), builder)
```
ChampionAsh5357, line 155: `map` is not a command on `Set`, but on `Stream`.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+    }
+    public Collection<String> getExamples() { // To provide examples shown in the chat on wrong command writing.
+        return List.of("Volvo", "Citroën");
+    }
+}
+```
+
+You will also need to add some functions to allow instantiation and get argument:
+ - A `public static CarArgumentType car()` that returns a `new CarArgumentType()` (it's a best pratice to avoid direct instanciations).
```
ChampionAsh5357, line 164: Why?

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+ - A function to get our arguments:
+ ```java
+ public static Car getCar(CommandContext<CommandSourceStack> ctx, String name) {
+    return ctx.getArgument(name, Car.class);
+ }
+ ```
```
ChampionAsh5357, line 170: Do you mean the method to get the provided user input for an argument key?

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+
+:::tip
+If you just wanna have custom suggestions like for an action (add, set, sub...) you can directly use `#suggests` after a `then` (that use a `RequiredArgumentType`) with as argument an equivalent of `CarAgumentType::listSuggestion`
+Another options is to create an enum with all the possibilities and use `EnumArgument.enumArgument(Cars.class)`.
+:::
+
+### `ArgumentTypeInfo`
+This class allow to serialize argument and provide build context.
```
ChampionAsh5357, line 178: The point of the info is to sync any data that is stored when constructing the argument type to the client (only for server commands).

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+public static class Info implements ArgumentTypeInfo<CarArgumentType, Template> {
+
+    public void serializeToNetwork(Template template, FriendlyByteBuf buffer) { // To serialize our argument into binary
+        buffer.writeByte(0); /* In this case, we don't have an ArgumentType argument, but if the constructor of 
+        CarArgumentType did have an argument, such as "boolean heavy", we would write something like
+        buffer.writeByte(template.heavy ? 1 : 0); */
+    }
+    public Template deserializeFromNetwork(FriendlyByteBuf buffer) { // To deserialize from the byte we wrote above
+        return new Template(); /* Taking an argument heavy like above, we have 
+        return new Template(buffer.readByte() == 1); */
```
ChampionAsh5357, line 192: Don't write the to the buffer and then not read the value.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+    }
+    public void serializeToJson(Template template, JsonObject json) { // To serialize into a JSON object
+        /* There's nothings to write here if new Template haven't any argument.
+        Otherwise, you'll write somethings like: json.addProperty("heavy", template.heavy); */
+    }
```
ChampionAsh5357, line 197: This is here for the commands report data provider.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+    public Template unpack(CarArgumentType arg) { // To transform the argument type instance into a Template<ArgumentType>
+        return new Template(); /* Or with an argument: new Template(arg.heavy); */
+    }
+
+    public static final class Template implements ArgumentTypeInfo.Template<CarArgmentType> {
```
ChampionAsh5357, line 202: This cannot be static for the reasons I explained in one of the other messages.

docs/misc/commands.md
```diff
@@ -0,0 +1,211 @@
...
+        public CarArgumentType instanciate() { // Says to Minecraft how to instanciate our argument
```
ChampionAsh5357, line 203: This is not the method to implement.

docs/misc/commands.md
CCCStudioCoder: done

docs/misc/commands.md
CCCStudioCoder: i talk over source code

Issue #182, opened 2024-10-23
Have you considered adding multi-language modules?


baka4n:
Is there any interest in having people from more countries translate this document

PR #272, opened 2025-08-30
I18n

CCCStudioCoder:
Create the foundation of the translation system.
It seems that there's a problem in the project structure.
Can somebody help me?


ChampionAsh5357:
As mentioned in #182, we currently have no plans to support other languages for the reasons stated.

PR #285, opened 2025-09-24
Fix category links shifting content on hover

Erdragh:
Making border width change conditionally shifts the contents of a website around. This fixes it for the categories on the root page by making the border width constant and only animating the border color.


sciwhiz12 (approved):
Simple and easy fix. Thanks!

sciwhiz12:
(We're on merge freeze at the moment in anticipation of 1.21.9, so please bear with us.)

Issue #293, opened 2025-10-07
Transfer APIs


ChampionAsh5357:
Add the documentation on the new Transfer APIs as while the docs do mention them, they never go into the underlying specifics.

PR #303, opened 2025-11-15
Add new Inventories & Transfers category and move some articles around

IchHabeHunger54:
This PR:
- Adds a new category named Inventories & Transfers, ordered between the Resources and Data Storage categories. It contains the following articles in order:
  - Containers (moved from the Block Entities category)
  - Capabilities (moved from the Data Storage category)
  - Menus (moved from the GUI category)
  - Transactions (to be written, see #293 )
- Moves the Screens article into the Rendering category, effectively removing the GUI category.
- In the Containers article, adjusts the introduction to make it less dependent on being in the Block Entities category.
- The changes are also applied to all versions back until 1.21.1, except for moving the Menus and Screens articles (due to lack of a Rendering category to move Screens into instead on those versions).

------------------
Preview URL: https://pr-303.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/blockentities/index.md
+++ b/docs/blockentities/index.md
@@ -267,7 +267,7 @@ It is important that you do safety checks, as the `BlockEntity` might already be
 [block]: ../blocks/index.md
 [blockreg]: ../blocks/index.md#basic-blocks
 [blockstate]: ../blocks/states.md
-[container]: container.md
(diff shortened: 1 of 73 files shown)


ChampionAsh5357 (approved):
The category name is a good length, so I agree.

PR #323, opened 2026-03-20
Include information on 26.1 versioning

cassiancc:
Rewrites the versioning page to include information on 26.1 versioning. I've also opted to remove some references to semver in relation to Minecraft, as the devs have repeatedly confirmed they do not follow it (and it's pretty obvious). 

Open to suggestions on structure for this, I had considered splitting the Minecraft heading in two for the two versioning phases, which would improve how the sidebar is structured.

Fixes #314 

------------------
Preview URL: https://pr-323.neoforged-docs-previews.pages.dev

PR #324, opened 2026-03-26
Update to 26.1

### LINKED ISSUE #326: ModifyDefaultComponentsEvent example needs updating
https://github.com/neoforged/NeoForge/pull/3041

ModifyDefaultComponentsEvent example needs updating as a breaking change was done for 26.1 to make it safe to now check for previous default components by passing those in to the event as a new parameter

### LINKED ISSUE #327: Typo for ModContainer in config docs
https://docs.neoforged.net/docs/misc/config/#registering-a-configuration
`ModConatiner#registerConfig`

### LINKED ISSUE #328: Container in Menus missing link.
shown `[Container][container]`

at https://docs.neoforged.net/docs/inventories/menus/#slot

### LINKED ISSUE #329: Deferred register docs use `BuiltInRegistries.BLOCKS`, which doesn't exist.
Deferred register docs use `BuiltInRegistries.BLOCKS`, which doesn't exist. The correct field is `BLOCK`.
https://docs.neoforged.net/docs/concepts/registries#methods-for-registering

### LINKED ISSUE #333: Entity render state modifiers
The example for modifiers on entity renderers without generics in https://docs.neoforged.net/docs/entities/renderer#render-state-modifications references the `PlayerRenderer` class which no longer exists. The reference needs to be replaced with another class (Neo's test mods use `PigRenderer` instead) and an example should be added for `RegisterRenderStateModifiersEvent#registerAvatarEntityModifier()` provided for bypassing issues with the intersection types in used in `AvatarRenderer`'s generics.

ChampionAsh5357:
Applies the contribution formatting and updates the docs to 26.1.

### Addressed

Closes #326 
Closes #327 
Closes #328 
Closes #329 
Supersedes #331 
Closes #333 
Supersedes #334 

### Subsections
 
- [x] Advanced
- [x] Block Entities
- [x] Blocks
- [x] Concepts
- [x] Data Storage
- [x] Entities
- [x] Getting Started (also by #323)
- [x] Inventories
- [x] Items
- [x] Misc
- [x] Networking
- [x] Rendering
- [x] Resources
    - [x] Client
        - [x] Models
    - [x] Server
        - [x] Data Maps
        - [x] Enchantments
        - [x] Loot Tables
        - [x] Recipes
- [x] World Generation

------------------
Preview URL: https://pr-324.neoforged-docs-previews.pages.dev


docs/misc/debugprofiler.md
```diff
@@ -7,7 +7,7 @@ Minecraft provides a Debug Profiler that provides system data, current game sett
 The Debug Profiler is very simple to use. It requires the debug keybind `F3 + L` to start the profiler. After 10 seconds, it will automatically stop; however, it can be stopped earlier by pressing the keybind again.
 
 :::note
-Naturally, you can only profile code paths that are actually being reached. `Entities` and `BlockEntities` that you want to profile must exist in the level to show up in the results.
+Naturally, you can only profile code paths that are actually being reached. `Entity`s and `BlockEntity`s that you want to profile must exist in the level to show up in the results.
```
IchHabeHunger54: Not sure I like this change. `Entity` and `BlockEntity` are the exact class names and should be kept as such imo. Also, while you're at it, link to the entity and block entity articles.

docs/items/datacomponents.md
```diff
@@ -214,12 +220,15 @@ public static final Item COMPONENT_EXAMPLE = REGISTRAR.register("component",
...
+        // Passes in the direct component value.
         .component(BASIC_EXAMPLE.get(), new ExampleRecord(24, true))
+        // Passes in a component factory, taking in the registry context and returning the value.
+        .delayedComponent(DataComponents.DAMAGE_RESISTANT, context -> new DamageResistant(context.getOrThrow(DamageTypeTags.IS_EXPLOSION)))
     )
 );
 ```
 
-If the data component should be added to an existing item that belongs to Vanilla or another mod, then `ModifyDefaultComponentsEvent` should be listened for on the [**mod event bus**][modbus]. The event provides the `modify` and `modifyMatching` methods which allows the `DataComponentPatch.Builder` to be modified for the associated items. The builder can either `#set` components or `#remove` existing components.
+If the data component should be added to an existing item that belongs to Vanilla or another mod, then `ModifyDefaultComponentsEvent` should be listened for on the [**mod event bus**][modbus]. The event provides the `modify` and `modifyMatching` methods which allows the `DataComponentPatch.Builder` to be modified for the associated items. The builder can either `#set` or remove existing components.
```
IchHabeHunger54:
```suggestion
If the data component should be added to an existing item that belongs to Vanilla or another mod, then `ModifyDefaultComponentsEvent` should be listened for on the [**mod event bus**][modbus]. The event provides the `modify` and `modifyMatching` methods which allows the `DataComponentPatch.Builder` to be modified for the associated items. The builder can `#set` existing components, including `#set`ting them to null which effectively removes them.
```

docs/items/index.md
```diff
@@ -169,9 +169,15 @@ However, this can sometimes lead to issues when dealing with multiple `ItemStack
 When in doubt, better be safe than sorry and `#copy` the stack.
 :::
 
+## `ItemStackTemplate`s
+
+`ItemStackTemplate`s are the immutable form of `ItemStack`s, typically representing a stack within an immutable context, such as recipes. Templates contain the basic elements that make up an `ItemStack`: the held holder `Item`, the stack size, and the [data components][datacomponents] the item has, stored as a patch.
+
+To create a new `ItemStackTemplate`, call one of the `new ItemTemplate(...)` methods, passing in the `Item` and any other desired elements. Then, when a stack is needed, an `ItemStack` can be created via `ItemStackTemplate#create`.
```
IchHabeHunger54:
```suggestion
To create a new `ItemStackTemplate`, call one of the `new ItemStackTemplate(...)` methods, passing in the `Item` and any other desired elements. Then, when a stack is needed, an `ItemStack` can be created via `ItemStackTemplate#create`.
```

docs/items/index.md
```diff
@@ -186,6 +192,12 @@ In many situations, for example [recipes], item stacks need to be represented as
 }
 ```
 
+## `ItemInstance`
+
+`ItemInstance` is a superinterface to which `ItemStack` and `ItemStackTemplate` implement. Generally, `ItemStack` and `ItemStackTemplate`s are used in isolated contexts. However, when the stack and template can be used interchangeably (e.g. the number of items in the stack / template), the `ItemInstance` superinterface is provided instead of a specific type.
```
IchHabeHunger54:
```suggestion
`ItemInstance` is a superinterface that `ItemStack` and `ItemStackTemplate` implement. Generally, `ItemStack` and `ItemStackTemplate`s are used in isolated contexts. However, when the stack and template can be used interchangeably (e.g. the number of items in the stack / template), the `ItemInstance` superinterface is provided instead of a specific type.
```

docs/items/index.md
```diff
@@ -186,6 +192,12 @@ In many situations, for example [recipes], item stacks need to be represented as
...
+
+`ItemInstance` provides common methods for checking the `Item` (`#s`), the stack size (`count`), and reading the data components through the `DataComponentGetter`.
```
IchHabeHunger54: `#s` just sounds wrong, is it `#stack`?

docs/rendering/screens.md
```diff
@@ -551,34 +551,34 @@ Since screens are subtypes of `GuiEventListener`s, the input handlers can also b
 
 ### Rendering the Screen
 
-Screens submit their elements through `#renderWithTooltipAndSubtitles` in three different strata: the background stratum, the render stratum, and the optional tooltip stratum.
+Screens submit their elements for rendering through `#extractRenderStateWithTooltipAndSubtitles` in three different strata: the background stratum, the element stratum, and the optional hoverable stratums.
```
IchHabeHunger54:
```suggestion
Screens submit their elements for rendering through `#extractRenderStateWithTooltipAndSubtitles` in three different strata: the background stratum, the element stratum, and the optional hoverable strata.
```

docs/rendering/screens.md
```diff
@@ -551,34 +551,34 @@ Since screens are subtypes of `GuiEventListener`s, the input handlers can also b
...
+Blurring, as handled through `GuiGraphicsExtractor#blurBeforeThisStratum`, can only be called once on any given frame. Attempting to submit a second blur will cause an exception to be thrown.
 :::
 
-The render stratum elements are submitted next via the `#render` method, provided by being a `Renderable` subtype. This mainly submits widgets and labels, along with setting the tooltips to render in the next stratum.
+The element stratum elements are submitted next via the `#extractRenderState` method, provided by being a `Renderable` subtype. This mainly submits widgets and labels, along with setting the hoverables to submit.
 
-Finally, the tooltip stratum submits the set tooltip. Tooltips are submitted in the render stratum via `GuiGraphics#setTooltipForNextFrame` or `GuiGraphics#setComponentTooltipFromElementsForNextFrame`, which can take in the text or tooltip components being submitted and the XY relative coordinates on where the tooltip should be rendered on the screen.
+Finally, the hoverable stratums submits things that hover over the previous elements, such as tooltips.
```
IchHabeHunger54:
```suggestion
Finally, the hoverable strata submit elements that hover over the previous elements, such as tooltips.
```

docs/rendering/screens.md
```diff
@@ -671,24 +673,9 @@ protected void containerTick() {
 
 ### Rendering the Container Screen
 
-The container screen uses all three strata to submit its elements. First, the background stratum submits the background texture via `#renderBg`. Then, the render stratum submits the widgets like before via `#render` followed by any text via `#renderLabels`. Finally, `AbstractContainerScreen` also provides the helper method `renderTooltip` to submit the tooltip to the tooltip stratum.
+The container screen uses all three strata to submit its elements. First, the background stratum submits the background texture by overriding `#extractBackground`. Then, the element stratum submits the widgets like before within `#extractContents`, followed by labels in `#extractLabels`. Finally, `AbstractContainerScreen` sets up the tooltip to be submitted during the hoverable stratums via `extractTooltip`
```
IchHabeHunger54:
```suggestion
The container screen uses all three strata to submit its elements. First, the background stratum submits the background texture by overriding `#extractBackground`. Then, the element stratum submits the widgets like before within `#extractContents`, followed by labels in `#extractLabels`. Finally, `AbstractContainerScreen` sets up the tooltip to be submitted during the hoverable strata via `extractTooltip`
```

docs/resources/client/models/index.md
```diff
@@ -22,11 +22,13 @@ A model is a JSON file with the following optional properties in the root tag:
...
-    - Block models should additionally specify a `particle` texture. This texture is used when falling on, running across, or breaking the block.
-    - Item models can also use layer textures, named `layer0`, `layer1`, etc., where layers with a higher index are rendered above those with a lower index (e.g. `layer1` would be rendered above `layer0`). Only works if the parent is `item/generated`, and only works for up to 5 layers (`layer0` through `layer4`).
+- `textures`: A sub-object that maps names (known as material variables) to `Material`s. Material variables can then be used in [elements]. They can also be specified in elements, but left unspecified in order for child models to specify them. Block models should additionally specify a `particle` texture. This texture is used when falling on, running across, or breaking the block. Item models can also use layer textures, named `layer0`, `layer1`, etc., where layers with a higher index are rendered above those with a lower index (e.g. `layer1` would be rendered above `layer0`). Only works if the parent is `item/generated`, and only works for up to 5 layers (`layer0` through `layer4`).
```
IchHabeHunger54: Is there a reason the two sub list entries for block particles and item layers have been collapsed into here?

docs/resources/client/models/items.md
```diff
@@ -1130,11 +1222,11 @@ public record ExampleSpecialRenderer(MaterialSet materialSet, Model.Simple model
 
     // Submit the features of the model
     @Override
-    public void submit(ItemDisplayContext displayContext, PoseStack poseStack, SubmitNodeCollector collector, int lightCoords, int overlayCoords, boolean hasFoil, int outlineColor) {
+    public void submit(Boolean argument, PoseStack poseStack, SubmitNodeCollector collector, int lightCoords, int overlayCoords, boolean hasFoil, int outlineColor) {
```
IchHabeHunger54, line 1225: Is it possible to use a `boolean` instead of a `Boolean` here?

docs/resources/server/recipes/builtin.md
```diff
@@ -174,41 +238,115 @@ TransmuteRecipeBuilder.transmute(RecipeCategory.MISC, this.tag(ItemTags.SHULKER_
...
+    },
+    "target": "minecraft:leather_boots"
+}
+```
+
+Like before, let's digest this line for line:
+
+- `type`: This is the id of the recipe serializer, `minecraft:crafting_transmute`.
+- `category`: This optional field defines the category in the crafting book.
+- `group`: This optional string property creates a group in the recipe book. Recipes in the same group will be displayed as one in the recipe book, which typically makes sense for transmuted recipes.
```
IchHabeHunger54:
```suggestion
- `group`: This optional string property creates a group in the recipe book. Recipes in the same group will be displayed as one in the recipe book, which typically makes sense for dying recipes.
```

docs/resources/server/recipes/builtin.md
```diff
@@ -174,41 +238,115 @@ TransmuteRecipeBuilder.transmute(RecipeCategory.MISC, this.tag(ItemTags.SHULKER_
         .save(this.output);
 ```
 
-### Special Crafting
+Additionally, `#addMaterialCountToOutput` can be used to add the number of materials used to the result count.
+
+### Dying Items
+
+Dye recipes are a special type of single item crafting recipes where the dye stacks' `DataComponents#DYE` are combined into a single `DataComponents#DYED_COLOR` and applied to the target stack to construct the result stack. All other components from the target stack are copied to the result stack, similar to [transmute crafting][transmute].  For example:
```
IchHabeHunger54:
```suggestion
Dye recipes are a special type of single item crafting recipes where the dye stacks' `DataComponents#DYE` are combined into a single `DataComponents#DYED_COLOR` and applied to the target stack to construct the result stack. All other components from the target stack are copied to the result stack, similar to [transmute crafting][transmute]. For example:
```

docs/resources/server/recipes/builtin.md
```diff
@@ -174,41 +238,115 @@ TransmuteRecipeBuilder.transmute(RecipeCategory.MISC, this.tag(ItemTags.SHULKER_
...
+    // Stores the recipe in the passed RecipeOutput, to be written to disk.
+    // If you want to add conditions to the recipe, those can be set on the output.
+    .save(this.output, "dyed_leather_boots");
+```
+
+### Special Recipes (TODO, rewrite)
+
+There are many other crafting recipes that are specifically made for a single purpose (e.g. cloning books, creating fireworks) rather than be applied in a generic manner. Most of the time, this is to set data components on the output by calculating their values from the input stacks. These recipes are still configurable, usually taking in the input ingredients and the template result. For example:
 
 In some cases, outputs must be created dynamically from inputs. Most of the time, this is to set data components on the output by calculating their values from the input stacks. These recipes usually only specify the type and hardcode everything else. For example:
```
IchHabeHunger54: This paragraph should probably be removed in favor of the one above.

docs/resources/server/recipes/custom.md
```diff
@@ -82,12 +85,111 @@ public class RightClickBlockRecipe implements Recipe<RightClickBlockInput> {
 }
 ```
 
-## Recipe Book Categories
+### Common Information
+
+All recipes have common information which, although may not be implemented the same way, is parsed from the JSON. For all recipes, vanilla provides `Recipe.CommonInfo`. Currently, it allows the recipe to specify whether to show the notification toast or not when unlocking. `CommonInfo` also provides a [map codec][codec] and [stream codec][streamcodec] for integrating with the [`RecipeSerializer`][serializer] below.
```
IchHabeHunger54:
```suggestion
All recipes have common information which, although it may not be implemented the same way, is parsed from the JSON. For all recipes, vanilla provides `Recipe.CommonInfo`. Currently, it allows the recipe to specify whether to show the notification toast or not when unlocking. `CommonInfo` also provides a [map codec][codec] and [stream codec][streamcodec] for integrating with the [`RecipeSerializer`][serializer] below.
```

IchHabeHunger54 (changes requested):

docs/misc/debugprofiler.md
cassiancc: This change makes it use the exact class names?

docs/resources/server/recipes/builtin.md
cassiancc:
```suggestion
- `group`: This optional string property creates a group in the recipe book. Recipes in the same group will be displayed as one in the recipe book, which typically makes sense for dyeing recipes.
```

docs/items/index.md
cassiancc:
```suggestion
To create a new `ItemStack`, call `new ItemStack(Item)`, passing in the backing item. By default, this uses a count of 1 and no NBT data; there are constructor overloads that accept a count and NBT data as well if needed. Note that an `ItemStack` cannot exist until components are bound/until a level exists. Until then, you should use an `ItemStackTemplate` as detailed below.
```

docs/items/armor.md
```diff
@@ -381,13 +381,15 @@ By default, the following layers render the associated `EquipmentClientInfo.Laye
 |:-----------------------:|:----------------------:|:---------------------------------------------------------------|
 | `HUMANOID`              | `HumanoidArmorLayer`   | Player, humanoid mobs (e.g., zombies, skeletons), armor stands |
 | `HUMANOID_LEGGINGS`     | `HumanoidArmorLayer`   | Player, humanoid mobs (e.g., zombies, skeletons), armor stands |
+| `HUMANOID_BABY`         | `HumanoidArmorLayer`   | Player, baby humanoid mobs (e.g. baby zombies), armor stands   |
```
cassiancc: Do armor stands use this at all? I believe they just scale down the `HUMANOID` model still.

cassiancc (changes requested):

docs/items/armor.md
ChampionAsh5357: Ah you're right. I don't really understand why `HumanoidArmorLayer` special cases it though since all other entities would use the humanoid baby armor, meaning they can't be viewed on an armor stand.

docs/items/index.md
ChampionAsh5357: It's supposed to be `is`

docs/resources/client/models/items.md
ChampionAsh5357: Nope since the `Boolean` is passed as part of the generic, and value types can't be generic.

docs/resources/server/recipes/builtin.md
ChampionAsh5357: Yep, just me leaving leftovers when I was rewriting as references.

docs/misc/debugprofiler.md
ChampionAsh5357: ^ That's the reason I changed it.

docs/resources/client/models/index.md
ChampionAsh5357: Since textures are now materials that have their own fields, they needed to be represented. I thought it would be confusing if the examples were inner layered with the fields on the textures. Should I pop this out into its own section instead?

docs/resources/index.md
```diff
@@ -54,18 +54,20 @@ Data packs may contain folders with files affecting the following things:
...
+| `instrument`, `jukebox_song`                                                                                              | Sound reference metadata     |
 | `painting_variant`                                                                                                        | Paintings                    |
 | `loot_table`                                                                                                              | [Loot tables][loottables]    |
 | `recipe`                                                                                                                  | [Recipes][recipes]           |
 | `tags`                                                                                                                    | [Tags][tags]                 |
 | `test_environment`, `test_instance`                                                                                       | [Game tests][gmt]            |
 | `trial_spawner`                                                                                                           | Combat challenges            |
```
cassiancc, line 69:
```suggestion
| `trade_set`                                                                                                                   | Villager trade sets            |
| `trial_spawner`                                                                                                           | Combat challenges            |
```

docs/resources/server/recipes/custom.md
```diff
@@ -395,12 +508,16 @@ Since recipe serializers can get fairly large, vanilla moves them to separate cl
 // are available, which were omitted from the code above.
 public class RightClickBlockRecipeSerializer implements RecipeSerializer<RightClickBlockRecipe> {
```
cassiancc: This is incorrect, see https://github.com/neoforged/Documentation/pull/334

docs/resources/client/models/index.md
```diff
@@ -153,21 +137,51 @@ Finally, a `neoforge:definition_type` can specify a custom model loader to regis
 
 ## Tinting
 
-Some blocks, such as grass or leaves, change their texture color based on their location and/or properties. [Model elements][elements] can specify a tint index on their faces, which will allow a color handler to handle the respective faces. The code side of things works through three events, one for block color handlers, one for block tints based on biome (used in conjunction with the block color handlers), and one for item tint sources. They work pretty similar, so let's have a look at a block handler first:
+Some blocks, such as grass or leaves, change their texture color based on their location and/or properties. [Model elements][elements] can specify a tint index on their faces, which will allow a color handler to handle the respective faces. The code side of things works through three events, one for block tint sources, one for block tints based on biome (used in conjunction with the block tint sources), and one for item tint sources. So let's have a look at a block tint source first:
 
 ```java
 @SubscribeEvent // on the mod event bus only on the physical client
 public static void registerBlockColorHandlers(RegisterColorHandlersEvent.Block event) {
```
cassiancc:
```suggestion
public static void registerBlockColorHandlers(RegisterColorHandlersEvent.BlockTintSources event) {
```

cassiancc (changes requested):

docs/items/armor.md
cassiancc: I believe the logic is that small armor stands don't have their own model, they're still scaled down armor stands. Having a way to view humanoid baby armour would be handy, but I'm not Mojang.

docs/resources/server/recipes/custom.md
ChampionAsh5357: I'm just gonna cherry-pick this

cassiancc (approved):

IchHabeHunger54 (approved):

PR #325, opened 2026-03-28
Document extra options for mixins in neoforge.mods.toml

lukebemish:
Currently there are several options in the mods.toml for mixins that are not documented here. One of them in particular -- `requiredMods` -- takes the place of something that many mods currently use mixin config plugins for, so it would be good to actually have docs on this feature.

------------------
Preview URL: https://pr-325.neoforged-docs-previews.pages.dev


ChampionAsh5357:
Adding this as a note for backporting:

- `behaviorVersion` was added as part of the `21.9` release
- `requiredMods` was added as part of `21.4.65-beta`

docs/gettingstarted/modfiles.md
```diff
@@ -174,6 +174,8 @@ public class ModTwo {
 | Property |  Type  |    Default    |             Description                       |     Example                       |
 |:--------:|:------:|:-------------:|:---------------------------------------------:|:----------------------------------|
 | `config` | string | **mandatory** | The location of the mixin configuration file. | `config="examplemod.mixins.json"` |
+| `requiredMods` | array | `[]` | Mod IDs that must be present for these mixins to be applied. | `requiredMods=["sodium"]` |
+| `behaviorVersion` | string | _nothing_ | Mixin may be upgraded after the end of a breaking change window. To avoid breaking mods, any new behavioral changes in mixin are gated behind a new behavior version, so if you need to rely on a behavior changed after the breaking change window, you can explicitly mark what mixin version you want the behavior of here. | `behaviorVersion="0.17.1"` |
```
ChampionAsh5357: So, what you are specifically trying to say here is that `behaviorVersion` represents the version of mixin required to run the mod. Is there any issues with this being less than the version currently on neo (e.g. neo on 0.16.5 but the mod defines 0.16.0)? Also, this should probably point out that we're using versions from https://github.com/FabricMC/Mixin along with where to find the current default version (e.g. `MixinFacade#DEFAULT_BEHAVIOUR_VERSION`) since it's not particularly clear.

ChampionAsh5357 (changes requested):

docs/gettingstarted/modfiles.md
lukebemish: It can be less than the version of mixin currently on neo, but it cannot be less than the version on neo when neo last left a BC window. I'll try to clarify a bit.

ChampionAsh5357 (approved):
Thanks!

PR #330, opened 2026-04-05
Fix Plural in mods section.

Exer0s:
Fixed a plural in the Mods section, version -> versions

### DIFF
--- a/docs/gettingstarted/versioning.md
+++ b/docs/gettingstarted/versioning.md
@@ -28,7 +28,7 @@ A few places in NeoForge also use [Maven version ranges][mvr], for example the M
 
 ## Mods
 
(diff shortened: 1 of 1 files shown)


ChampionAsh5357:
So, I don't think that the error with this paragraph. It's saying that there is no on best versioning system, which imo is correct. However, there is an error with the sentence after that, replacing the system there with systems since it's talking about multiple being combined.

If you plan on making that fix, please rebase on the `26.1` branch and update all versions.

Exer0s:
> So, I don't think that the error with this paragraph. It's saying that there is no on best versioning system, which imo is correct. However, there is an error with the sentence after that, replacing the system there with systems since it's talking about multiple being combined.
> 
> If you plan on making that fix, please rebase on the `26.1` branch and update all versions.

Wow, I messed up that easy! That was the instance of the word I meant to change, not sure how I did that. Will rebase and close this one, whoops!

PR #334, opened 2026-04-13
[26.1] Fix recipe serializer docs

cassiancc:
As `RecipeSerializer` is now a record, not an interface, it can no longer be implemented directly and must have codecs passed into the constructor.

------------------
Preview URL: https://pr-334.neoforged-docs-previews.pages.dev


cassiancc:
Cherry-picked into #324.

PR #342, opened 2026-04-30
Fluids

### LINKED ISSUE #109: Add docs for fluids

IchHabeHunger54:
Closes #109 .

Adds fluid documentation. Pretty big one, and also based off my one (recent) time of implementing a fluid myself, so I'm not 100% sure I caught and correctly understood everything.

**Edit:** Since the article is getting bigger than anticipated, I decided to create an entire new fluids category.

------------------
Preview URL: https://pr-342.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/blocks/fluids.md
+++ b/docs/blocks/fluids.md
@@ -1,0 +1,151 @@
+---
+description: How to work with fluids, fluid states and fluid stacks, and how to add your own.
+sidebar_position: 3
+---
+# Fluids
+
+In vanilla Minecraft, the two fluids - water and lava - are special types of [blocks][block] that can spread to neighboring blocks over a certain distance. They are generally not solid, and [entities][entity] can enter and "swim" in them.
+
+In modded Minecraft, especially in many tech mods, fluids also take on the role of recipe ingredients. This is possible because fluids exist in a separate registry and are only added to the world using fluid blocks, essentially meaning that fluids can be seen in complete independence from blocks.
+
+This article aims to showcase both the in-world and the recipe aspects of fluids.
+
+:::warning
+Due to vanilla only having two fluids, and those fluids having a lot of special-casing, some of these systems are very hacky and - due to a lot of edge cases that cannot be reasonably caught in testing - may not always work correctly. If you find a bug with fluids, please reach out to us on Discord.
+:::
+
+## `Fluid` and `FluidType`
+
+Before we can register a fluid, we must first understand a few design decisions made by Minecraft and NeoForge.
+
+In Minecraft, water and lava each have two variants: a flowing fluid and a source fluid. The way this works is mostly due to hardcoding, in some association with `FluidState`s (see below). Since this hardcoding is inconvenient at best and practically impossible to use at worst, NeoForge introduces the `FluidType` class and patches a ton of places to use it. The main purpose of the `FluidType` is to contain the common logic of the fluid - e.g. the sounds it makes, whether boats can be used in it, etc. - and only leave the actual flowing logic in the fluid itself. `FluidType`s live in a separate registry added by NeoForge, and thus must be registered in addition to `Fluid`s.
+
+With that in mind, let's start creating our fluid! For the sake of example, we're going to create a molten iron fluid. To get started, we need two [registries][registries]:
+
+```java
+public static final DeferredRegister<Fluid> FLUIDS =
+    DeferredRegister.create(Registries.FLUID, ExampleMod.MOD_ID);
+public static final DeferredRegister<FluidType> FLUID_TYPES =
+    DeferredRegister.create(NeoForgeRegistries.FLUID_TYPES, ExampleMod.MOD_ID);
+```
+
+Since `Fluid`s require a `FluidType` to be created, we create the `FluidType` first. A `FluidType`'s options are defined in a `Properties` object, similar to block properties.
+
+```java
+public static final DeferredHolder<FluidType, FluidType> MOLTEN_IRON_TYPE = FLUID_TYPES.register(
+        // The registry name of the fluid type. Usually it makes sense to name it the same as the `Fluid`.
+        "molten_iron",
+        // The supplier for the fluid type, accepting a `FluidType.Properties` object.
+        () -> new FluidType(FluidType.Properties.create()
+                // The translation key of the fluid. While this will not be visible in vanilla Minecraft,
+                // it will be visible if the fluid is stored in e.g. a modded tank, or when looked at in-world
+                // with WAILA (What Am I Looking At?) or similar mods installed.
+                // In order to later make datagen easier, we use a block translation key here.
+                // If you do not plan on adding a block, you can replace "block." with "fluid."
+                .descriptionId("block." + ExampleMod.MOD_ID + ".molten_iron")
+                // Set lava-like sounds for our fluid. This is only relevant if you have a bucket item,
+                // which we will look at later.
+                .sound(SoundActions.BUCKET_FILL, SoundEvents.BUCKET_FILL_LAVA)
+                .sound(SoundActions.BUCKET_EMPTY, SoundEvents.BUCKET_EMPTY_LAVA)
+                // We cannot swim or drown in molten iron.
+                .canDrown(false)
+                .canSwim(false)
+                // We want molten iron to slightly glow.
+                .lightLevel(5)
+        ));
+```
+
+:::tip
+There are a bunch of other methods in `FluidType`. For example, if you were to make a more water-like fluid, the `supportsBoating()` and `isWaterLike()` methods could be interesting to you. For a full list of available methods, please see the source of `FluidType.Properties`.
+
+Not all of these methods are used by vanilla systems. Some of them, such as `temperature()` or `density()`, were requested in the original design phase of the `FluidType` system for mod compatibility, and may or may not be used by modded systems.
+:::
+
+With our `FluidType` created, we can move to the `Fluid` itself. NeoForge provides the `BaseFlowingFluid` class as a base for us to use, which has three inner classes: `Source`, `Flowing` and `Properties`. `Source` and `Flowing` are subclasses of `BaseFlowingFluid`, following the layout of vanilla's `WaterFluid` and `LavaFluid`, while `Properties` is once again a block properties-like object, this time responsible for tying the fluid type, source fluid, flowing fluid and later also stuff like the bucket or the fluid block together.
+
+Since the source and flowing fluids depend on the fluid properties but the fluid properties also depends on the two fluids, we need to be a little careful with static initialization order and qualify with the class name in some places. Assuming you are keeping your fluids in a class named `ModFluids`, the code looks as follows:
+
+```java
+// The source fluid. This is usually named without specifying "source" in the name. 
+public static final DeferredHolder<Fluid, BaseFlowingFluid.Source> MOLTEN_IRON = FLUIDS.register(
+        // The registry name.
+        "molten_iron",
+        // The source fluid supplier. Qualify the properties with the class name here.
+        () -> new BaseFlowingFluid.Source(ModFluids.MOLTEN_IRON_PROPERTIES));
+
+// The flowing fluid. The name is commonly prefixed with "flowing_".
+public static final DeferredHolder<Fluid, BaseFlowingFluid.Flowing> FLOWING_MOLTEN_IRON = FLUIDS.register(
+        // The registry name.
+        "flowing_molten_iron",
+        // The flowing fluid supplier. Again, qualify the properties with the class name.
+        () -> new BaseFlowingFluid.Flowing(ModFluids.MOLTEN_IRON_PROPERTIES));
+
+// The fluid properties. We will use this later to connect additional stuff to the fluid, for example the bucket.
+public static final BaseFlowingFluid.Properties MOLTEN_IRON_PROPERTIES =
+        // Parameters are the fluid type, the source fluid and the flowing fluid.
+        new BaseFlowingFluid.Properties(MOLTEN_IRON_TYPE, MOLTEN_IRON, FLOWING_MOLTEN_IRON);
+```
+
+With this done, your fluid should now be loaded into the game, and recipes will be able to make use of it.
+
+## Resources
+
+While our fluid now exists, we aren't done yet: we still need to add the resource files for the fluid. For a fluid without a block, this is limited to textures and a translation. Blocks later also require a model and a renderer to be set up.
+
+Let's start by adding the texture files. When creating your assets, it is recommended to use the vanilla water or lava texture as a basis; this is especially important with flowing fluids as they use what is effectively a 2x2 texture that is sampled by the flowing fluid renderer. The texture files must be named and placed as follows (where `examplemod` is your mod id):
+
+- `assets/examplemod/textures/block/molten_iron_still.png` for the still texture, and
+- `assets/examplemod/textures/block/molten_iron_flowing.png` for the flowing texture.
+
+Most fluids are animated, so they will also need accompanying `.png.mcmeta` files. Again, you can base these off the vanilla files. For more information, see the article on [textures].
+
+Now for the translations. The translation key used by fluids is defined by `FluidType#descriptionId()`. In our example, we used `block.examplemod.molten_iron`, so we would add a translation like so:
+
+```java
+    @Override
+    protected void addTranslations() {
+        // other translations here
+    
+        add("block.examplemod.molten_iron", "Molten Iron");
+        
+        // Alternatively, once you have created a fluid block later:
+        addBlock(ModBlocks.MOLTEN_IRON.get(), "Molten Iron");
+    }
+```
+
+For more information, see [I18n and L10n/Datagen][i18n].
+
+## In-World Fluids
+
+TODO
+
+### `FluidState` and Waterlogging
+
+TODO
+
+### Fluid Blocks
+
+TODO
+
+### Cauldrons
+
+TODO
+
+## Fluids in Recipes
+
+TODO
+
+### `FluidStack`
+
+TODO
+
+### `FluidIngredient`
+
+TODO
+
+[block]: index.md
+[entity]: ../entities/index.md
+[i18n]: ../resources/client/i18n.md#datagen
+[registries]: ../concepts/registries.md
+[tags]: ../resources/server/tags.md#datagen
+[textures]: ../resources/client/textures.md
--- a/docs/blocks/fluids.md
+++ b/docs/blocks/fluids.md
@@ -117,11 +117,49 @@ For more information, see [I18n and L10n/Datagen][i18n].
 
 ## In-World Fluids
 
-TODO
+When placing fluids in world, `FluidState`s are used instead of `Fluid`s, closely mirroring the use of [`BlockState`s][blockstate] versus `Block`s. Similar to `BlockState`s, `FluidState`s can be set into a level using `Level#setFluidState()`, a `FluidState` at a position can be queried using `Level#getFluidState()`, and the default state can be obtained using `Fluid#defaultFluidState()`.
 
-### `FluidState` and Waterlogging
+However, `FluidState`s also exhibit a few differences to `BlockState`s. Most notably, their different states do not operate using properties, at least not properties defined in the same way as block state properties, instead the exact `FluidState` is computed by the level from fluid spreading mechanics. For most use cases the exact `FluidState` is irrelevant, save for some properties such as `isSource()` which can be queried from the `FluidState` if needed.
 
-TODO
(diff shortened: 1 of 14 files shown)


docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+# Fluids
+
+In vanilla Minecraft, the two fluids - water and lava - are special types of [blocks][block] that can spread to neighboring blocks over a certain distance. They are generally not solid, and [entities][entity] can enter and "swim" in them.
+
+In modded Minecraft, especially in many tech mods, fluids also take on the role of recipe ingredients. This is possible because fluids exist in a separate registry and are only added to the world using fluid blocks, essentially meaning that fluids can be seen in complete independence from blocks.
+
+This article aims to showcase both the in-world and the recipe aspects of fluids.
+
+:::warning
+Due to vanilla only having two fluids, and those fluids having a lot of special-casing, some of these systems are very hacky and - due to a lot of edge cases that cannot be reasonably caught in testing - may not always work correctly. If you find a bug with fluids, please reach out to us on Discord.
```
ChampionAsh5357: Or... you know, create an issue on the Github.

docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+        // The registry name of the fluid type. Usually it makes sense to name it the same as the `Fluid`.
+        "molten_iron",
+        // The supplier for the fluid type, accepting a `FluidType.Properties` object.
+        () -> new FluidType(FluidType.Properties.create()
+                // The translation key of the fluid. While this will not be visible in vanilla Minecraft,
+                // it will be visible if the fluid is stored in e.g. a modded tank, or when looked at in-world
+                // with WAILA (What Am I Looking At?) or similar mods installed.
+                // In order to later make datagen easier, we use a block translation key here.
+                // If you do not plan on adding a block, you can replace "block." with "fluid."
+                .descriptionId("block." + ExampleMod.MOD_ID + ".molten_iron")
```
ChampionAsh5357: `Util.makeDescriptionId` would make it better. Then, you can use the `Identifier` passed into factory rather than the supplier variant.

docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+With this done, your fluid should now be loaded into the game, and recipes will be able to make use of it.
+
+## Resources
+
+While our fluid now exists, we aren't done yet: we still need to add the resource files for the fluid. For a fluid without a block, this is limited to textures and a translation. Blocks later also require a model and a renderer to be set up.
+
+Let's start by adding the texture files. When creating your assets, it is recommended to use the vanilla water or lava texture as a basis; this is especially important with flowing fluids as they use what is effectively a 2x2 texture that is sampled by the flowing fluid renderer. The texture files must be named and placed as follows (where `examplemod` is your mod id):
+
+- `assets/examplemod/textures/block/molten_iron_still.png` for the still texture, and
+- `assets/examplemod/textures/block/molten_iron_flowing.png` for the flowing texture.
```
ChampionAsh5357: The location is determined by the registered `FluidModel` in `RegisterFluidModelsEvent`

docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+Most fluids are animated, so they will also need accompanying `.png.mcmeta` files. Again, you can base these off the vanilla files. For more information, see the article on [textures].
+
+Now for the translations. The translation key used by fluids is defined by `FluidType#descriptionId()`. In our example, we used `block.examplemod.molten_iron`, so we would add a translation like so:
+
+```java
+    @Override
+    protected void addTranslations() {
+        // other translations here
+    
+        add("block.examplemod.molten_iron", "Molten Iron");
```
ChampionAsh5357: You literally just mentioned using `FluidType#getDescriptionId`, so why didn't you use it here?

docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+While our fluid now exists, we aren't done yet: we still need to add the resource files for the fluid. For a fluid without a block, this is limited to textures and a translation. Blocks later also require a model and a renderer to be set up.
+
+Let's start by adding the texture files. When creating your assets, it is recommended to use the vanilla water or lava texture as a basis; this is especially important with flowing fluids as they use what is effectively a 2x2 texture that is sampled by the flowing fluid renderer. The texture files must be named and placed as follows (where `examplemod` is your mod id):
+
+- `assets/examplemod/textures/block/molten_iron_still.png` for the still texture, and
+- `assets/examplemod/textures/block/molten_iron_flowing.png` for the flowing texture.
+
+Most fluids are animated, so they will also need accompanying `.png.mcmeta` files. Again, you can base these off the vanilla files. For more information, see the article on [textures].
+
+Now for the translations. The translation key used by fluids is defined by `FluidType#descriptionId()`. In our example, we used `block.examplemod.molten_iron`, so we would add a translation like so:
```
ChampionAsh5357:
```suggestion
Now for the translations. The translation key used by fluids is defined by `FluidType#getDescriptionId()`. In our example, we used `block.examplemod.molten_iron`, so we would add a translation like so:
```

docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+        add("block.examplemod.molten_iron", "Molten Iron");
+        
+        // Alternatively, once you have created a fluid block later:
+        addBlock(ModBlocks.MOLTEN_IRON.get(), "Molten Iron");
+    }
+```
+
+For more information, see [I18n and L10n/Datagen][i18n].
+
+## In-World Fluids
```
ChampionAsh5357: I feel like this section probably needs to be rewritten a bit since the way it's framed doesn't give a clear picture. First, there is no 'setting a fluid state' since the fluid state is directly tied to the block state. The only way to set a fluid state is to replace the liquid block or set some kind of liquid-logging property, which is specific to the implementing block.

docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+
+Unfortunately, the current implementation of `FluidState`s in levels is very much half-baked. Even more unfortunately, it is impossible for NeoForge to fix this without breaking compatibility with vanilla worlds. Basically all `FluidState` logic is tied to `BlockState` in some way, despite there not really being a need to. In the current implementation, `Level#getFluidState()` essentially boils down to `BlockState#getFluidState()`, happening very deep in chunk storage. It is expected that Mojang will eventually rework this, however for now we have to make do with what we have.
+
+### Waterlogging
+
+_See also [Blocks][block] and [Block States][blockstate]._
+
+The epitome of this half-baked `FluidState` system is waterlogging. Waterlogging is the ability of certain non-full blocks, e.g. slabs, to also contain a water source at the same time. This is currently implemented via the `WATERLOGGED` block state property:
```
ChampionAsh5357: This doesn't have to be water, that is just the vanilla usecase.

docs/blocks/fluids.md
```diff
@@ -0,0 +1,190 @@
...
+    // The `false` parameter in Fluids.WATER.getSource(false) means "falling" and is set to false
+    // for all vanilla waterlogging implementations.
+    @Override
+    public FluidState getFluidState(BlockState state) {
+        return state.getValue(WATERLOGGED) ? Fluids.WATER.getSource(false) : super.getFluidState(state);
+    }
+}
+```
+
+The `WATERLOGGED` is also the #1 reason NeoForge cannot easily fix this, because removing the `WATERLOGGED` property would break compatibility with vanilla servers due to a different set of block states.
```
ChampionAsh5357: I don't think this needs to be justified. It doesn't really help the understanding of what waterlogging is. Docs are supposed to be informative and not a justification piece on why the system is not good.

ChampionAsh5357 (changes requested):

docs/blocks/fluids.md
IchHabeHunger54: I like to tell people to reach out on Discord first, to gauge whether this is a new issue, a known issue, or WAI. But I will add both.

docs/fluids/index.md
```diff
@@ -0,0 +1,200 @@
...
+        // The registry name.
+        "flowing_molten_iron",
+        // The flowing fluid supplier. Again, qualify the properties with the class name.
+        () -> new BaseFlowingFluid.Flowing(ModFluids.MOLTEN_IRON_PROPERTIES));
+
+// The fluid properties. We will use this later to connect additional stuff
+// to the fluid, for example the bucket.
+public static final BaseFlowingFluid.Properties MOLTEN_IRON_PROPERTIES =
+        // Parameters are the fluid type, the source fluid and the flowing fluid.
+        new BaseFlowingFluid.Properties(MOLTEN_IRON_TYPE, MOLTEN_IRON, FLOWING_MOLTEN_IRON);
```
ChampionAsh5357, line 92: I wonder if we should also show a vanilla example since it's not guaranteed they'll use `BaseFlowingSource` or split flowing from source.

docs/fluids/index.md
```diff
@@ -0,0 +1,200 @@
...
+            // if it is null, no overlay will be displayed. Overlays are only used for in-world fluids,
+            // so if you don't have an in-world fluid, it should always be null.
+            new Material(Identifier.fromNamespaceAndPath(ExampleMod.MOD_ID, "block/molten_iron_still")),
+            new Material(Identifier.fromNamespaceAndPath(ExampleMod.MOD_ID, "block/molten_iron_flowing")),
+            new Material(Identifier.fromNamespaceAndPath(ExampleMod.MOD_ID, "block/molten_iron_overlay")),
+            // The fluid tint source. We leave it at null, which means no tint. See below for more info.
+            null),
+            // Suppliers for the still and flowing fluids.
+            ModFluids.MOLTEN_IRON::value,
+            ModFluids.FLOWING_MOLTEN_IRON::value
```
ChampionAsh5357: Did you forget that `DeferredHolder` is a supplier?

docs/fluids/index.md
```diff
@@ -0,0 +1,200 @@
...
+    );
+}
+```
+
+## Resources
+
+While our fluid now exists, we aren't done yet: we still need to add the resource files for the fluid. This consists of textures, a model and a translation.
+
+Let's start by adding the texture files. When creating your assets, it is recommended to use the vanilla water or lava texture as a basis; this is especially important with flowing fluids as they use what is effectively a 2x2 texture that is sampled by the flowing fluid renderer. The texture files must be named and placed as follows (where `examplemod` is your mod id):
```
ChampionAsh5357, line 123: It should probably be made clear that fluid textures must live in the block atlas, hence why they are in the `block` subdirectory.

docs/fluids/index.md
```diff
@@ -0,0 +1,200 @@
...
+Most fluids are animated, so they will also need accompanying `.png.mcmeta` files. Again, you can base these off the vanilla files. For more information, see the article on [textures].
+
+
+Finally, the translations. The translation key used by fluids is defined by `FluidType#descriptionId()`, and we can get it from a `FluidType` using `#getDescriptionId()`:
+
+```java
+@Override
+protected void addTranslations() {
+    add(AMFluids.MOLTEN_IRON_TYPE.getDescriptionId(), "Molten Iron");
+}
```
ChampionAsh5357, line 144: Where is this from? Need to mention that this is within a `LanguageProvider`

docs/fluids/index.md
```diff
@@ -0,0 +1,200 @@
...
+// If possible, we want to use a singleton.
+public final class MoltenIronTintSource implements FluidTintSource {
+    public static final MoltenIronTintSource INSTANCE = new MoltenIronTintSource();
+    
+    private MoltenIronTintSource() {}
+
+    @Override
+    public int color(FluidState state) {
+        // Return whatever color you want here.
+        return 0xff000000;
```
ChampionAsh5357, line 166: Color is in ARGB format. If no alpha is specified, then the rendering will be invisible iirc.

docs/fluids/index.md
```diff
@@ -0,0 +1,200 @@
...
+```java
+@Override
+protected void addTranslations() {
+    add(AMFluids.MOLTEN_IRON_TYPE.getDescriptionId(), "Molten Iron");
+}
+```
+
+For more information, see [I18n and L10n/Datagen][i18n].
+
+## Fluid Tint Sources
```
ChampionAsh5357, line 149: Where is the fluid custom renderer gonna go, since it's also passed into the fluid model.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+## Waterlogging
+
+_See also [Blocks][block] and [Block States][blockstate]._
+
+The epitome of the half-baked `FluidState` system is waterlogging. Waterlogging is the ability of certain non-full blocks, e.g. slabs, to also contain a water source at the same time. This is currently implemented via the `WATERLOGGED` block state property:
+
+```java
+// Implementing SimpleWaterloggedBlock automatically enables bucket pickup
+// and makes some helper methods available.
+public class MyBlock extends Block implements SimpleWaterloggedBlock {
```
ChampionAsh5357, line 22: This is missing some behavior. The two obvious ones are scheduling the fluid tick when the shape is updated, and setting the state for placement if placing in a fluid-filled tile. For a simplish teardown, look at `BarrierBlock`.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+    // for all vanilla waterlogging implementations.
+    @Override
+    public FluidState getFluidState(BlockState state) {
+        return state.getValue(WATERLOGGED) ? Fluids.WATER.getSource(false) : super.getFluidState(state);
+    }
+}
+```
+
+:::info
+"Lavalogging" or similar fluid-logging with other fluids is easily possible. To do so, simply create a new `BooleanProperty`, add it to the block as usual, and have `Block#getFluidState()` return the desired fluid if the property is true.
```
ChampionAsh5357, line 85: There is a bit more than that since there are some hardcoded instances that checks for the waterlogged property, but for the general 'this block only' behavior, yes.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+Fluids can usually be picked up in a bucket. A custom bucket for our fluid can be added like so:
+
+```java
+// Assuming a DeferredRegister.Items named ITEMS, and assuming the fluid stuff
+// is in another class named ModFluids.
+public static final DeferredItem<BucketItem> MOLTEN_IRON_BUCKET = ITEMS.registerItem(
+        // The registry name.
+        "molten_iron_bucket",
+        // The bucket item factory.
+        properties -> new BucketItem(AMFluids.LIQUID_ETHERIUM.get(), properties),
```
ChampionAsh5357: Ah yes, liquid etherium.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+```
+
+Next, it is recommended (but not required) to add a dispenser behavior for the bucket:
+
+```java
+@SubscribeEvent // on the mod event bus
+private static void commonSetup(FMLCommonSetupEvent event) {
+    // `DispenserBlock#registerBehavior` is not thread-safe so we wrap it in a lambda.
+    // The anonymous class seen here is copied from `DispenseItemBehavior#bootStrap()`.
+    event.enqueueWork(() -> DispenserBlock.registerBehavior(ModItems.MOLTEN_IRON_BUCKET, new DefaultDispenseItemBehavior() {
```
ChampionAsh5357: Sure, though there's also neo's `DispenseFluidContainer`.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+:::
+
+Finally, all that's left is a translation and a model:
+
+```java
+// In the language provider
+@Override
+protected void addTranslations() {
+    add(AMFluids.MOLTEN_IRON_TYPE.get().getDescriptionId(), "Molten Iron");
+    addItem(AMItems.MOLTEN_IRON_BUCKET, "Molten Iron Bucket");
```
ChampionAsh5357: Hi `AM`, how're you doing today? Gonna ignore the other instances and just mention, polish.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+protected void registerModels(BlockModelGenerators blockModels, ItemModelGenerators itemModels) {
+    blockModels.createNonTemplateModelBlock(ModBlocks.MOLTEN_IRON.get());
+    // We use NeoForge's `DynamicFluidContainerModel`.
+    itemModels.itemModelOutput.accept(AMItems.LIQUID_ETHERIUM_BUCKET.get(), new DynamicFluidContainerModel.Unbaked(
+        // The model's textures.
+        new DynamicFluidContainerModel.Textures(
+                Optional.of(new Material(Identifier.withDefaultNamespace("item/bucket"))),
+                Optional.of(new Material(Identifier.withDefaultNamespace("item/bucket"))),
+                Optional.of(new Material(Identifier.fromNamespaceAndPath("neoforge", "item/mask/bucket_fluid"))),
+                Optional.empty()
```
ChampionAsh5357, line 205: I feel like this could use more comments.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+        ),
+        // The fluid to use.
+        AMFluids.LIQUID_ETHERIUM.get(),
+        // Whether the bucket model should be flipped, commonly used for "gaseous" fluids.
+        false,
+        // If true, the "cover" texture is a mask. We generally want this for buckets.
+        true,
```
ChampionAsh5357: What is a mask? (I know what it is, but explain).

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+```java
+@SubscribeEvent // on the mod event bus
+private static void registerCauldronFluidContent(RegisterCauldronFluidContentEvent event) {
+    event.register(
+            // The cauldron block.
+            ModBlocks.MOLTEN_IRON_CAULDRON.get(),
+            // The fluid.
+            ModFluids.MOLTEN_IRON.get(),
+            // The amount. 1000 is one bucket.
+            1000,
```
ChampionAsh5357: `FluidType#BUCKET_VOLUME`.

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+}
+```
+
+Secondly, we need to register the actual interactions. That works like so:
+
+```java
+@SubscribeEvent
+private static void registerCauldronInteractions(RegisterCauldronInteractionEvent.Interaction event) {
+    // Empty our cauldron when it is right-clicked with an empty bucket.
+    MoltenIronCauldronBlock.CAULDRON_INTERACTIONS.put(Items.BUCKET,
```
ChampionAsh5357: Shouldn't we be using `register` instead of directly calling the interaction map to add?

docs/fluids/inworld.md
```diff
@@ -0,0 +1,383 @@
...
+            // The resulting block state.
+            ModBlocks.MOLTEN_IRON_CAULDRON.get().defaultBlockState(),
+            // The sound event to play when filling the cauldron.
+            SoundEvents.BUCKET_EMPTY_LAVA));
+}
+```
+
+Cauldron interactions are not limited to buckets. Vanilla adds a couple of other cauldron recipes, mostly for "cleaning" colored items. These work through generally the same mechanism. For more information, see the `CauldronInteractions` class. This is also where you can find the vanilla cauldron interaction dispatchers.
+
+[block]: ../blocks/index.md
```
ChampionAsh5357, line 432: What happened to the `FluidInteractionRegistry`?

ChampionAsh5357 (changes requested):

docs/fluids/inworld.md
IchHabeHunger54: I'm not actually 100% sure, so please let me know if I messed up the explanation.

docs/fluids/inworld.md
IchHabeHunger54: As far as I know it's currently defunct?

ChampionAsh5357: ?? It's still used in `LiquidBlock#neighborChanged`.

docs/fluids/inworld.md
ChampionAsh5357: When `coverIsMask` is true, we are treating the cover texture as a mask for the base texture. More specifically, for each pixel where the cover texture has an alpha of 0, the cover sprite (as long as the base sprite is present) is shown, and where the cover texture has a non-zero alpha, the base sprite is shown.

This really isn't necessary for buckets unless you want your buckets to have a cover. What you're thinking of is the fluid mask most likely. After all, the cover only provides a mask to the base texture (e.g., the bucket container), not the fluid. Of course, there are reasons to mask the base texture, such as if you don't want the fluid to be tinted by the base texture if transparency is involved, though in those cases, `force_opaque_fluid` is typically set to true instead.

docs/fluids/inworld.md
IchHabeHunger54: According to other team members on the discord, in 26.1 fluids (and fluid interactions in particular) are considerably fucked at the moment. `FluidInteractionRegistry` is very due for a rework and the patches are broken. I would suggest we update the docs if and when it gets fixed.

docs/fluids/index.md
IchHabeHunger54: Addressed in an info block near the top.

PR #343, opened 2026-04-30
Fix formatting in armor.md documentation

WBattist:


ChampionAsh5357:
Nautili is the correct plural of nautilus.

WBattist:
‘Nautili’ is the traditional plural rooted in Latin, while ‘nautiluses’ is also correct and more common in modern English. This doesn’t really matter, and I already committed the wrong term, but I just wanted to point it out.

PR #344, opened 2026-05-03
[26.1] Transfer API

### LINKED ISSUE #293: Transfer APIs

ChampionAsh5357:
Documents the transfer API.

Closes #293 

------------------
Preview URL: https://pr-344.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/inventories/capabilities.md
+++ b/docs/inventories/capabilities.md
@@ -32,14 +32,14 @@ Here is an example of discouraged capability usage:
 
 NeoForge provides capabilities for the following three [resource handlers][resourcehandler]: `ResourceHandler<ItemResource>`, `ResourceHandler<FluidResource>` and `EnergyHandler`.
 
-`ResourceHandler<ItemResource>` exposes a resource handler for handling inventory slots. The capabilities of type `ResourceHandler<ItemResource>` are:
+`ResourceHandler<ItemResource>` exposes an interface for managing inventory slots. The capabilities of type `ResourceHandler<ItemResource>` are:
 
 - `Capabilities.Item.BLOCK`: automation-accessible inventory of a block (for chests, machines, etc).
 - `Capabilities.Item.ENTITY`: inventory contents of an entity (extra player slots, mob/creature inventories/bags).
 - `Capabilities.Item.ENTITY_AUTOMATION`: automation-accessible inventory of an entity (boats, minecarts, etc).
 - `Capabilities.Item.ITEM`: contents of an item stack (portable backpacks and such).
 
-`ResourceHandler<FluidResource>` exposes a resource handler for handling fluid inventories. The capabilities of type `ResourceHandler<FluidResource>` are:
+`ResourceHandler<FluidResource>` exposes an interface for managing fluid inventories. The capabilities of type `ResourceHandler<FluidResource>` are:
 
 - `Capabilities.Fluid.BLOCK`: automation-accessible fluid inventory of a block.
 - `Capabilities.Fluid.ENTITY`: fluid inventory of an entity.
--- a/docs/inventories/capabilities.md
+++ b/docs/inventories/capabilities.md
@@ -30,22 +30,22 @@ Here is an example of discouraged capability usage:
 
 ## NeoForge-provided capabilities
 
-NeoForge provides capabilities for the following three interfaces: `ResourceHandler<ItemResource>`, `ResourceHandler<FluidResource>` and `EnergyHandler`.
+NeoForge provides capabilities for the following three [resource handlers][resourcehandler]: `ResourceHandler<ItemResource>`, `ResourceHandler<FluidResource>` and `EnergyHandler`.
 
-`ResourceHandler<ItemResource>` exposes an interface for handling inventory slots. The capabilities of type `ResourceHandler<ItemResource>` are:
+`ResourceHandler<ItemResource>` exposes a resource handler for handling inventory slots. The capabilities of type `ResourceHandler<ItemResource>` are:
 
 - `Capabilities.Item.BLOCK`: automation-accessible inventory of a block (for chests, machines, etc).
 - `Capabilities.Item.ENTITY`: inventory contents of an entity (extra player slots, mob/creature inventories/bags).
 - `Capabilities.Item.ENTITY_AUTOMATION`: automation-accessible inventory of an entity (boats, minecarts, etc).
 - `Capabilities.Item.ITEM`: contents of an item stack (portable backpacks and such).
 
-`ResourceHandler<FluidResource>` exposes an interface for handling fluid inventories. The capabilities of type `ResourceHandler<FluidResource>` are:
+`ResourceHandler<FluidResource>` exposes a resource handler for handling fluid inventories. The capabilities of type `ResourceHandler<FluidResource>` are:
 
 - `Capabilities.Fluid.BLOCK`: automation-accessible fluid inventory of a block.
 - `Capabilities.Fluid.ENTITY`: fluid inventory of an entity.
 - `Capabilities.Fluid.ITEM`: fluid inventory of an item stack.
 
-`EnergyHandler` exposes an interface for handling energy containers. It is based on the RedstoneFlux API by TeamCoFH. The capabilities of type `EnergyHandler` are:
+`EnergyHandler` exposes an interface for handling energy containers. It is based on the RedstoneFlux API by TeamCoFH. The capabilities of type [`EnergyHandler`][energyhandler] are:
 
 - `Capabilities.Energy.BLOCK`: energy contained inside a block.
 - `Capabilities.Energy.ENTITY`: energy containing inside an entity.
@@ -62,7 +62,7 @@ Capabilities allow looking up implementations of some APIs with some dispatching
 - `EntityCapability`: capabilities for entities: behavior depends on the specific `EntityType`.
     - The capability commonly specifies a `Direction` context for different resources depending on the side.
 - `ItemCapability`: capabilities for item stacks: behavior depends on the specific `Item`.
-    - The capability commonly specifies an `ItemAccess` context for the holding item resource.
+    - The capability commonly specifies an [`ItemAccess`][itemaccess] context for the holding item resource.
 
 :::tip
 For compatibility with other mods, we recommend using the capabilities provided by NeoForge in the `Capabilities` class if possible. Otherwise, you can create your own as described in this section.
@@ -347,4 +347,7 @@ See [`CapabilityHooks`][capability-hooks] for a list of the providers registered
 
 [block-cap-provider]: https://github.com/neoforged/NeoForge/blob/26.1.x/src/main/java/net/neoforged/neoforge/capabilities/IBlockCapabilityProvider.java
 [capability-hooks]: https://github.com/neoforged/NeoForge/blob/26.1.x/src/main/java/net/neoforged/neoforge/capabilities/CapabilityHooks.java
+[energyhandler]: transactions.md#energy-handler
 [invalidation]: #block-capability-invalidation
+[itemaccess]: transactions.md#item-access
+[resourcehandler]: transactions.md#resource-handlers
--- a/docs/inventories/transactions.md
+++ b/docs/inventories/transactions.md
@@ -152,7 +152,7 @@ int amountInserted = handler.insert(0, ItemResource.of(Items.APPLE), 5, ctx);
 int amountExtracted = handler.extract(ItemResource.of(Items.APPLE), 2, ctx);
 ```
 
-There are many different types of `ResourceHandler`s depending on what the backing inventory is. Some handlers wrap around existing vanilla inventories (e.g., `VanillaContainerWrapper` for [`Container`s][container], `PlayerInventoryWrapper` for the [player's `Inventory`][playerinv], `LivingEntityEquipmentWrapper` for a [living entity's][livingentity] equipment slots).
+There are many different types of `ResourceHandler`s depending on what the backing inventory is. Some handlers wrap around existing vanilla inventories (e.g., `VanillaContainerWrapper` for [`Container`s][container], `PlayerInventoryWrapper` for [player `Inventory`s][playerinv], `LivingEntityEquipmentWrapper` for a [living entity's][livingentity] equipment slots).
 
 ```java
 // Wrapping around an existing inventory.
--- a/docs/inventories/transactions.md
+++ b/docs/inventories/transactions.md
@@ -1,6 +1,6 @@
 # Transactions
 
-Transactions are a NeoForged-added system for managing the communication between different inventories transferring their contents. Each transfer is managed through three basic concepts: the `Resource`s being transferred, the `ResourceHandler`s representing the inventory, and the `Transaction` facilitating the communication.
+Transactions are a NeoForged-added system for managing the communication between different inventories transferring their contents. Each transfer is managed through three basic concepts: the `Resource`s being transferred, the `ResourceHandler`s representing the inventories, and the `Transaction` facilitating the communication.
 
 ## Resources
 
@@ -369,9 +369,9 @@ NeoForge also provides a `ResourceStacksResourceHandler`, using `ResourceStack`s
 
 ### Energy Handler
 
-`EnergyHandler` is a trimmed down version of `ResourceHandler`, only containing one index storing a `long`. As such, it only checks how many units can be stored (`getCapacityAsLong` / `getCapacityAsInt`) along with how many units already stored (`getAmountAsLong` / `getAmountAsInt`). Additionally, `insert` and `extract` no longer take in an index since there's only one, and a `Resource` as the backing object is a primitive.
+`EnergyHandler` is a trimmed down version of `ResourceHandler`, only containing one index storing a `long`. As such, it only checks how many units can be stored (`getCapacityAsLong` / `getCapacityAsInt`) along with how many units already stored (`getAmountAsLong` / `getAmountAsInt`). Additionally, `insert` and `extract` no longer take in an index since there's only one, and also no longer require a `Resource`, as the backing object is a primitive.
 
-Like `ResourceHandler`, there are different types of `EnergyHandler`s depending on usecase. The most common is `SimpleEnergyHandler`, which provides a basic implementation along with any limits on insert / extract.
+Like `ResourceHandler`, there are different types of `EnergyHandler`s depending on your use case. The most common one is `SimpleEnergyHandler`, which provides a basic implementation along with a limit for insert / extract.
 
 ```java
 // Create an energy handler.
@@ -496,7 +496,7 @@ public class ExampleResourceHandler extends SnapshotJournal<ExampleObject> imple
         // This should be immutable.
         ExampleObject original = this.object;
         this.object = new ExampleObject(
-            original.id(), original.count(), new HashMap<>(original.flags())
+            original.id(), original.count(), ImmutableMap.copyOf(original.flags())
         );
         return original;
     }
--- a/docs/inventories/transactions.md
+++ b/docs/inventories/transactions.md
@@ -1,0 +1,609 @@
+# Transactions
+
+Transactions are a NeoForged-added system for managing the communication between different inventories transferring their contents. Each transfer is managed through three basic concepts: the `Resource`s being transferred, the `ResourceHandler`s representing the inventory, and the `Transaction` facilitating the communication.
+
+## Resources
+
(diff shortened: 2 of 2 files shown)


docs/inventories/transactions.md
```diff
@@ -0,0 +1,609 @@
+# Transactions
+
+Transactions are a NeoForged-added system for managing the communication between different inventories transferring their contents. Each transfer is managed through three basic concepts: the `Resource`s being transferred, the `ResourceHandler`s representing the inventory, and the `Transaction` facilitating the communication.
+
+## Resources
+
+`Resource`s represent the backing object that is transacted upon. Each `Resource` is meant to be immutable, containing what the object is rather than how much is being transferred. For example, the transaction 'five apples for an emerald', contains the `Resource`s 'apple' and 'emerald', not 'five apples' and 'one emerald'.
```
IchHabeHunger54: I'd word this a little differently. Something like "containing only what kind of object is used, not the amount of objects"

docs/inventories/transactions.md
```diff
@@ -0,0 +1,609 @@
+# Transactions
+
+Transactions are a NeoForged-added system for managing the communication between different inventories transferring their contents. Each transfer is managed through three basic concepts: the `Resource`s being transferred, the `ResourceHandler`s representing the inventory, and the `Transaction` facilitating the communication.
```
IchHabeHunger54:
```suggestion
Transactions are a NeoForged-added system for managing the communication between different inventories transferring their contents. Each transfer is managed through three basic concepts: the `Resource`s being transferred, the `ResourceHandler`s representing the inventories, and the `Transaction` facilitating the communication.
```

docs/inventories/transactions.md
```diff
@@ -0,0 +1,609 @@
...
+int indexCapacity = handler.getCapacityAsInt(0);
+boolean canUseApples = handler.isValid(0, ItemResource.of(Items.APPLE));
+
+// Put a resource in at a specific index given some `TransactionContext` ctx.
+int amountInserted = handler.insert(0, ItemResource.of(Items.APPLE), 5, ctx);
+// Extract a resource given some `TransactionContext` ctx.
+int amountExtracted = handler.extract(ItemResource.of(Items.APPLE), 2, ctx);
+```
+
+There are many different types of `ResourceHandler`s depending on what the backing inventory is. Some handlers wrap around existing vanilla inventories (e.g., `VanillaContainerWrapper` for [`Container`s][container], `LivingEntityEquipmentWrapper` for a [living entity's][livingentity] equipment slots).
```
IchHabeHunger54: Is there a wrapper for the player inventory? If yes, it may be worth mentioning here.

docs/inventories/transactions.md
```diff
@@ -0,0 +1,609 @@
...
+}
+```
+
+:::tip
+NeoForge also provides a `ResourceStacksResourceHandler`, using `ResourceStack`s as the stored contents, for `Resource` implementations that are themselves the actual objects within an inventory.
+:::
+
+### Energy Handler
+
+`EnergyHandler` is a trimmed down version of `ResourceHandler`, only containing one index storing a `long`. As such, it only checks how many units can be stored (`getCapacityAsLong` / `getCapacityAsInt`) along with how many units already stored (`getAmountAsLong` / `getAmountAsInt`). Additionally, `insert` and `extract` no longer take in an index since there's only one, and a `Resource` as the backing object is a primitive.
```
IchHabeHunger54:
```suggestion
`EnergyHandler` is a trimmed down version of `ResourceHandler`, only containing one index storing a `long`. As such, it only checks how many units can be stored (`getCapacityAsLong` / `getCapacityAsInt`) along with how many units already stored (`getAmountAsLong` / `getAmountAsInt`). Additionally, `insert` and `extract` no longer take in an index since there's only one, and also no longer require a `Resource`, as the backing object is a primitive.
```

docs/inventories/transactions.md
```diff
@@ -0,0 +1,609 @@
...
+
+Like `ResourceHandler`, there are different types of `EnergyHandler`s depending on usecase. The most common is `SimpleEnergyHandler`, which provides a basic implementation along with any limits on insert / extract.
```
IchHabeHunger54:
```suggestion
Like `ResourceHandler`, there are different types of `EnergyHandler`s depending on your use case. The most common one is `SimpleEnergyHandler`, which provides a basic implementation along with a limit for insert / extract.
```

docs/inventories/transactions.md
```diff
@@ -0,0 +1,609 @@
...
+    
+    // ...
+
+    @Override
+    protected ExampleObject createSnapshot() {
+        // Create a snapshot of the object.
+        // This should be immutable.
+        ExampleObject original = this.object;
+        this.object = new ExampleObject(
+            original.id(), original.count(), new HashMap<>(original.flags())
```
IchHabeHunger54:
```suggestion
            original.id(), original.count(), ImmutableMap.copyOf(original.flags())
```
For proper immutability.

IchHabeHunger54 (changes requested):
Mostly a look across grammar and typos, leaving the contents to the ones who actually implemented the system.

IchHabeHunger54 (dismissed):
Approving from my end, but wait for the technical reviews before merging.

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
+# Transactions
```
Technici4n, line 1: This should be transfer, not transactions, no?

docs/inventories/capabilities.md
```diff
@@ -30,22 +30,22 @@ Here is an example of discouraged capability usage:
 
 ## NeoForge-provided capabilities
 
-NeoForge provides capabilities for the following three interfaces: `ResourceHandler<ItemResource>`, `ResourceHandler<FluidResource>` and `EnergyHandler`.
+NeoForge provides capabilities for the following three [resource handlers][resourcehandler]: `ResourceHandler<ItemResource>`, `ResourceHandler<FluidResource>` and `EnergyHandler`.
 
-`ResourceHandler<ItemResource>` exposes an interface for handling inventory slots. The capabilities of type `ResourceHandler<ItemResource>` are:
+`ResourceHandler<ItemResource>` exposes a resource handler for handling inventory slots. The capabilities of type `ResourceHandler<ItemResource>` are:
```
Technici4n: Not a fan of this: if you don't already know what a "resource handler" is (terrible name btw but such is life), then "a resource handler for handling" is not helpful

docs/inventories/transactions.md
Technici4n: It is important to note that transactions are only used for operations that modify an inventory. Reading an inventory for example does not require the use of transactions. Similarly, resources are a concept that goes beyond transactions or even inventories. For example `Set<ItemResource>` might be the cleanest way to represent the items allowed by a filter.

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+    }
+}
+
+// Create our resource.
+public final class ExampleResource implements Resource {
+
+    private final ExampleObject object;
+
+    public ExampleResource(ExampleObject object) {
+        // Enforce immutability and ignore count.
```
Technici4n, line 74: This is not the greatest example IMO, it's very specific to the case of wrapping a mutable stack. Ideally one would only deal in immutable resources (or at least consider it strongly when writing entirely new code).

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+    @Override
+    public String toString() {
+        return Integer.toString(this.object.id()) + "[" 
+            + this.object.flags().size() + "]";
+    }
+}
+```
+
+:::note
+While `Resource`s can be used for primitives, they are not strictly necessary (e.g., energy does not have a `Resource` as it is backed by a `long`). However, it does require reimplementing some of the resource behavior yourself, as the [handler system described below][handler] requires the use of a `Resource`.
```
Technici4n, line 128: What is a "primitive" here?

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+:::
+
+## Resource Handlers
+
+`ResourceHandler<T>`s represent the backing inventories within a transaction, where `T` is the type of the `Resource` backing the object. Each handler maps to its associated contents using an index (e.g., index `0` maps to slot 1, index `1` maps to slot 2, etc). For every index, you can check whether a `Resource` can be contained at the location (`isValid`) or what `Resource` is already stored there (`getResource`). You can also check how many `Resource`s can be stored at the location (`getCapacityAsLong` / `getCapacityAsInt`) along with how many of a `Resource` is stored there (`getAmountAsLong` / `getAmountAsInt`). The number of indices accessible to the handler represents its `size`.
```
Technici4n: > the backing inventories within a transaction

not really; you don't need a transaction to read from an inventory

Technici4n: > index `0` maps to slot 1, index `1` maps to slot 2, etc

this is IMO confusing; why mix 0-indexing and 1-indexing?

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+// For some ResourceHandler<ItemResource> handler
+
+// Get the resource stored in the handler.
+ItemResource item = handler.getResource(0);
+int count = handler.getAmountAsInt(0);
+
+// Get information about the handler itself.
+int handlerSize = handler.size();
+int indexCapacity = handler.getCapacityAsInt(0);
+boolean canUseApples = handler.isValid(0, ItemResource.of(Items.APPLE));
```
Technici4n: "use"?

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+
+// Put a resource in at a specific index given some `TransactionContext` ctx.
+int amountInserted = handler.insert(0, ItemResource.of(Items.APPLE), 5, ctx);
+// Extract a resource given some `TransactionContext` ctx.
+int amountExtracted = handler.extract(ItemResource.of(Items.APPLE), 2, ctx);
```
Technici4n: I would probably avoid showing I/O code without the `tx.commit()` and the wrapping `try` block.

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+```
+
+There are many different types of `ResourceHandler`s depending on what the backing inventory is. Some handlers wrap around existing vanilla inventories (e.g., `VanillaContainerWrapper` for [`Container`s][container], `PlayerInventoryWrapper` for [player `Inventory`s][playerinv], `LivingEntityEquipmentWrapper` for a [living entity's][livingentity] equipment slots).
+
+```java
+// Wrapping around an existing inventory.
```
Technici4n: container? ;)

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+If you plan to use one of the `StacksResourceHandler`s as an inventory, it is highly recommended to override `onContentsChanged` to handle any disk writing or network syncing.
+
+```java
+// Example for block entities
+public class ExampleBlockEntity extends BlockEntity {
+
+    private final ItemStacksResourceHandler storage = new ItemStacksResourceHandler(5) {
+        @Override
+        protected void onContentsChanged(int index, ItemStack previousContents) {
+            // Mark the block entity as dirty
```
Technici4n:
```suggestion
            // Schedule the block entity for saving
```
or similar... "mark dirty" is old terminology

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+
+// Wrapping around a specific equipment slot for some LivingEntity entity.
+ResourceHandler<ItemResource> head = LivingEntityEquipmentWrapper.of(entity, EquipmentSlot.HEAD);
+```
+
+Other handlers are themselves inventories, providing a convenience for those wanting to make use of the system without much implementing (e.g., `ItemStacksResourceHandler` for a list of [`ItemStack`s][itemstack], `FluidStacksResourceHandler` for a list of `FluidStack`s).
+
+```java
+// Creating an `ItemStack` storage.
+ItemStacksResourceHandler itemStorage = new ItemStacksResourceHandler(5);
```
Technici4n, line 168: This is likely the first thing that should be shown, as it's what we want people to use (equivalent to the old `ItemStackHandler`).

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+        TransferPreconditions.checkNonEmptyNonNegative(resource, amount);
+
+        // Check whether the resource can be inserted from this location.
+        ExampleObject current = this.object;
+        if (current.count() == 0 || (current.id() == resource.id() && current.flags().equals(resource.flags()) && this.isValid(index, resource))) {
+            // Compute the amount to insert.
+            int insertedAmount = Math.min(amount, this.getCapacityAsInt(index, resource) - current.count());
+
+            if (insertedAmount > 0) {
+                // Update the content.
```
Technici4n, line 266: This needs a call to `updateSnapshots(transaction)` (which also requires extending from `SnapshotJournal`) for correctness, otherwise the change will not be rolled back correctly. So I would maybe keep custom implementations for the end as a more advanced example.

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+        TransferPreconditions.checkNonEmptyNonNegative(resource, amount);
+
+        // Check whether the resource can be extracted from this location.
+        ExampleObject current = this.object;
+        if (current.id() == resource.id() && current.flags().equals(resource.flags())) {
+            // Compute the amount to extract.
+            int extracted = Math.min(current.count(), amount);
+
+            if (extracted > 0) {
+                // Update the content.
```
Technici4n, line 299: Same, this needs `updateSnapshots`.

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    @Override
+    public boolean matches(ExampleObject object, ExampleResource resource) {
+        // Check if an object matches the stored resource.
+        return object.id() == resource.id() && object.flags().equals(resource.flags());
+    }
+}
+```
+
```
Technici4n, line 363: This is missing a discussion of the `ResourceHandlerUtil` methods, which are very helpful for making efficient use of the API.

docs/inventories/transactions.md
```diff
@@ -0,0 +1,613 @@
...
+// Put energy in given some `TransactionContext` ctx.
+int amountInserted = energy.insert(100, ctx);
+// Take energy out given some `TransactionContext` ctx.
+int amountExtracted = energy.extract(10, ctx);
+```
+
+### Item Access
+
+`ItemAccess` is also a trimmed down version of `ResourceHandler`, providing access to a single item in a specific storage location. As such, it only provides the resource (`getResource`) and amount of the item currently present (`getAmount`). Additionally, `insert` and `extract` no longer take in an index since there's only one. However, as items can also store data, the `ItemAccess` provides a way to access the stored data in through [capabilities] via `getCapability`, assuming it is an `ItemCapability` with an `ItemAccess` context.
```
Technici4n: I guess it is a trimmed down version, but the key point is really that this is used by item capabilities to modify their item stack.

Technici4n (commented):
Overall not sure about the structure. It's difficult to introduce such a large API, but at the same time I would much rather have people copy/paste something complicated that they do not fully understand, than something simple but incorrect. It might be worth splitting into a few pages maybe?

For reference, I wrote these tutorials some years ago, and the APIs are almost the same: https://wiki.fabricmc.net/tutorial:transfer-api.

docs/inventories/capabilities.md
ChampionAsh5357: I'll go with `interface for managing inventory slots` then. That's the best way I can think of saying it.

docs/inventories/transactions.md
ChampionAsh5357: In my understanding, the whole purpose of the resource and resource handling was because of the transaction system. Without the transaction system, I do not really understand the benefit of using resources over vanilla implementations or the caps we had previously.

Also, transfer is a verb, which feels a bit awkward for a title. Same with 'transfer API', as I don't think that accurately describes what you would use these systems for. Hence, I decided to focus on transaction for the purpose of the piece and explaining it through that lens.

Though, we can always think of a better title or a way to frame this.

docs/inventories/transactions.md
ChampionAsh5357: I wanted the example to be transferrable to existing resources, since both the item and fluid implementations make use of a backing mutable object. I believe it gets the immutability portion across clearly. Though, if you have a better idea, please give it a whirl.

docs/inventories/transactions.md
ChampionAsh5357: Java primitive like an `int`, `long`, `float`, etc. This is my understanding after seeing that `EnergyHandler` does not have a backing resource.

docs/inventories/transactions.md
ChampionAsh5357: True, though this document was originally framed through the lens of a transaction.

ChampionAsh5357: I can say first slot, second slot instead. I just wanted to subtly hint at the mapping is just an index to an arbitrary slot. You could have the handler's index read the inventory in the reverse order if you so choose.

docs/inventories/transactions.md
ChampionAsh5357: Shh... should be accept.

docs/inventories/transactions.md
ChampionAsh5357: Fine with me, though then that really limits the general explanation of resource handling to the transaction lens as, although the specific implementations may have direct get and set methods, the base interface is built with transactions in mind.

docs/inventories/transactions.md
ChampionAsh5357: Look, all containers are inventories if you think about it. Not `Inventory` of course :p.

docs/inventories/transactions.md
ChampionAsh5357: I think we need to have more of a discussion here then since I believe this is less important here and more important in another inventory doc.

docs/inventories/transactions.md
ChampionAsh5357: Fair enough.

docs/inventories/transactions.md
ChampionAsh5357: I have that in the example below. The example here is just to provide an understanding of what the handler is doing. Then, we integrate it into the transaction context.

docs/inventories/transactions.md
ChampionAsh5357: Same as above where this is just an example of how it's implemented, with the snapshot stuff coming below.

docs/inventories/transactions.md
ChampionAsh5357: Probably because I forgot all about it.

docs/inventories/transactions.md
pupnewfster: > Without the transaction system, I do not really understand the benefit of using resources over vanilla implementations or the caps we had previously.

Yes, that is why resource handlers are intrinsically linked to using transactions for inserting/extracting from them (though querying contents without any claims about if they can be inserted/extracted). The main improvements resource handlers have over the old system or vanilla is: transactions, and being generic. I think what tech meant in terms of the example of transactions not being necessary for reading the inventory, is that if you have a case where you can accept a certain set of resources, you don't need to open a transaction when seeing if the resource stored in an index is valid for your set, and then if it is valid you can open it before actually trying to extract from it.

That being said resources themselves are useful for general handling when you don't necessarily care about (item vs fluid vs modded types) for a piece of logic. Such as tech's example of in a modded filter to store the types that match without also storing the unused "count" or having to duplicate the filter's implementation between items and fluids.

ChampionAsh5357:
> Overall not sure about the structure. It's difficult to introduce such a large API, but at the same time I would much rather have people copy/paste something complicated that they do not fully understand, than something simple but incorrect. It might be worth splitting into a few pages maybe?
> 
> For reference, I wrote these tutorials some years ago, and the APIs are almost the same: https://wiki.fabricmc.net/tutorial:transfer-api.

Probably, though I think it may need further discussion. Some of your suggestions talk about resources and handlers as their own separate concept while others refer to them within the transaction context. Personally, I don't mind examples, but the neo docs are explanatory at their core, meaning we need explain details than just provide a copy-paste tutorial.

In my understanding, `Resource`s and `ResourceHandler`s are grounded in the concept that you will be performing `Transaction`s with them. Without transactions, the base API for `Resource`s or `ResourceHandler`s don't really exist, making it just a more generic neo-implementation of vanilla containers.

We could have an additional page for the specific implementations or integrate into the container docs for the specific `ResoruceHandler`s, but I do not have an understanding of how you are picturing the Transfer API system to accurately document them.

Technici4n (approved):
Sure

Minecraftschurli (approved):

PR #357, opened 2026-05-30
Fix documentation for additionalRuntimeClasspath for <=1.21.9

c-Caelum:
When adding libraries via additionalRuntimeClasspath, you must exclude the dependencies and add them back manually. I believe this is because gradle sees it as having already been added, even though it does not handle the transitive dependencies.

------------------
Preview URL: https://pr-357.neoforged-docs-previews.pages.dev

### DIFF
--- a/toolchain/docs/dependencies/nonmclibs.md
+++ b/toolchain/docs/dependencies/nonmclibs.md
@@ -34,7 +34,7 @@ dependencies {
     additionalRuntimeClasspath 'com.example:example:1.0'
 }
 ```
-
+:::warning
 However, using a library that has transitive dependencies will require you to manually exclude and `additionalRuntimeClasspath` / `implementation` the dependency yourself. This is only needed if that specific module is required at runtime; compilation will proceed fine.
 
 ```gradle
@@ -43,13 +43,14 @@ dependencies {
     implementation 'com.example:example:1.0'
     // This adds the library to all the runs
     additionalRuntimeClasspath 'com.example:example:1.0' {
-        // exclude the module you want to add to the classpath
-        exclude 'com.example:my_module'
+        // exclude the transitive dependency you want to add to the classpath
+        exclude 'com.example:dependency'
     }
     // add the module back to the classpath
     additionalRuntimeClasspath 'com.example:my_module:1.0'
 }
 ```
+:::
 
 </TabItem>
 <TabItem value="ng" label="NeoGradle">
--- a/toolchain/docs/dependencies/nonmclibs.md
+++ b/toolchain/docs/dependencies/nonmclibs.md
@@ -35,7 +35,7 @@ dependencies {
 }
 ```
 :::warning
-However, using a library that has transitive dependencies will require you to manually exclude and `additionalRuntimeClasspath` / `implementation` the dependency yourself. This is only needed if that specific module is required at runtime; compilation will proceed fine.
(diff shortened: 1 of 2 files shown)


ChampionAsh5357:
Apologies for the late reply. I would stick this block in a `warning` tag and give a bit more prefix of the explanation that `example` depends on `my_module` transitively, meaning the implementation will add it as a compile-time dep. It may even be better to change `my_module` to something like `dependency` or something to make it more explicit.

c-Caelum:
Understood. I'll make those changes.

toolchain/docs/dependencies/nonmclibs.md
```diff
@@ -34,6 +34,23 @@ dependencies {
     additionalRuntimeClasspath 'com.example:example:1.0'
 }
 ```
+:::warning
+However, using a library that has transitive dependencies will require you to manually exclude and `additionalRuntimeClasspath` / `implementation` the dependency yourself. This is only needed if that specific module is required at runtime; compilation will proceed fine.
```
ChampionAsh5357:
```suggestion
If the library contains any transitive dependencies required at runtime, the transitive dependencies have to be manually excluded and readded to the `additionalRuntimeClasspath` configuration.
```

Reworded it. I took out the `implementation` since compile time should handle that without issue.

toolchain/docs/dependencies/nonmclibs.md
```diff
@@ -34,6 +34,23 @@ dependencies {
     additionalRuntimeClasspath 'com.example:example:1.0'
 }
 ```
+:::warning
```
ChampionAsh5357, line 37: There should be a newline before this.

toolchain/docs/dependencies/nonmclibs.md
```diff
@@ -34,6 +34,23 @@ dependencies {
...
+However, using a library that has transitive dependencies will require you to manually exclude and `additionalRuntimeClasspath` / `implementation` the dependency yourself. This is only needed if that specific module is required at runtime; compilation will proceed fine.
+
+```gradle
+dependencies {
+    // This is required to add the library at compile time
+    implementation 'com.example:example:1.0'
+    // This adds the library to all the runs
+    additionalRuntimeClasspath 'com.example:example:1.0' {
+        // exclude the transitive dependency you want to add to the classpath
```
ChampionAsh5357:
```suggestion
        // Exclude the transitive dependency that is required at runtime
```

Just being more explicit.

toolchain/docs/dependencies/nonmclibs.md
```diff
@@ -34,6 +34,23 @@ dependencies {
...
+        exclude 'com.example:dependency'
+    }
+    // add the module back to the classpath
+    additionalRuntimeClasspath 'com.example:my_module:1.0'
```
ChampionAsh5357:
```suggestion
    // Add the module back to the runtime classpath
    additionalRuntimeClasspath 'com.example:dependency:1.0'
```

Missed `my_module` change.

ChampionAsh5357 (changes requested):
Just a few more things, then all good.

c-Caelum:
Okay, I think that's done?

ChampionAsh5357:
Please remove the `package-lock.json` changes.

ChampionAsh5357:
Creating a commit that has a removal message does not revert the changes. Please revert the `package-lock.json` changes.

PR #368, opened 2026-07-27
Reload listeners

### LINKED ISSUE #306: Add docs for resource listeners

IchHabeHunger54:
Adds reload listener docs. Since this covers both client and server (as they are almost the same), they are placed in the root Resources folder.

~~Opening as draft as it is not yet complete. At the time of writing, the second half of the article, as well as any and all links, are missing.~~ **EDIT:** Done.

Closes #306 .

------------------
Preview URL: https://pr-368.neoforged-docs-previews.pages.dev

### DIFF
--- a/docs/resources/reloadlisteners.md
+++ b/docs/resources/reloadlisteners.md
@@ -3,13 +3,13 @@ sidebar_position: 3
 ---
 # Reload Listeners
 
-In some situations, integrating with the existing resource systems provided by Minecraft or NeoForge just isn't going to cut it. Instead, having your system load files by itself from a resource or data pack is more desirable. For this purpose, you can register a custom reload listener, implementing `PreparableReloadListener` or one of its subinterfaces/subclasses.
+In some situations, integrating with the [existing resource systems][resources] provided by Minecraft or NeoForge just isn't going to cut it. Instead, having your system load files by itself from a resource or data pack is more desirable. For this purpose, you can register a custom reload listener, implementing `PreparableReloadListener` or one of its subinterfaces/subclasses.
 
 The idea behind a reload listener is simple: When a resource pack or data pack reload happens, the listener is called upon to read its contents from the new set of resource or data packs. It will then keep the contents until the next reload, at which point the contents will be discarded and the cycle starts anew.
 
 ## Reloading
 
-Both resource pack and data pack reload function similar in principle, and only differ in timing and the location files are loaded from.
+Both resource pack and data pack reload function similar in principle and only differ in the associated [side][sides], and by extension the timing and the location files are loaded from.
 
 Client reload listeners load from resource packs (the `assets` folder). The first reload happens during startup on the physical client, and **never** on the physical server. Subsequent reloads are triggered when resource packs are changed in the Options menu, or by pressing F3+T.
 
@@ -23,7 +23,7 @@ Reloading happens on multiple threads, as reload listeners are unrelated to one 
 
 ## Adding a Reload Listener
 
-All reload listeners are registered using the same basic principle. First, we need a reload listener instance. Vanilla typically stores its reload listeners, such as the texture or recipe managers, as fields in `Minecraft` (for client-side listeners) or `ServerLevel` (for server-side listeners), however in modded contexts, a singleton instance is usually fine as well:
+All reload listeners are registered using the same basic principle. First, we need a reload listener instance. Vanilla typically stores its reload listeners, such as the [texture][textures] or [recipe managers][recipes], as fields in `Minecraft` (for client-side listeners) or `ServerLevel` (for server-side listeners), however in modded contexts, a singleton instance is usually fine as well:
 
 ```java
 // Instead of PreparableReloadListener, extend one of its subclasses if applicable, see below.
@@ -40,7 +40,7 @@ public class MyReloadListener implements PreparableReloadListener {
 }
 ```
 
-Then, depending on whether the reload listener is for client data (resource packs) or server data (data packs), the listener is registered to one of two events:
+Then, depending on whether the reload listener is for client data (resource packs) or server data (data packs), the listener is registered to one of two [events][events]:
 
 ```java
 // For client-side reload listeners
@@ -57,7 +57,7 @@ public static void addServerReloadListeners(AddServerReloadListenersEvent event)
 ```
 
 :::danger
-Do not register the same reload listener on both sides! All of your file-driven systems should be designed for one side only, otherwise behavior might differ between singleplayer and multiplayer.
+Do not register the same reload listener on both sides! All of your file-driven systems should be designed for one side only, otherwise desyncs and similar issues will arise.
 :::
 
 And then, the reload listener can be accessed - from the correct side - through the singleton `INSTANCE`.
@@ -112,7 +112,7 @@ And then simply access your values like so:
 MyObject myObject = MyReloadListener.INSTANCE.get(Identifier.fromNamespaceAndPath("mymod", "example"));
 ```
 
-Note that `SimpleJsonResourceReloadListener` goes through the resource packs top to bottom, and only retains the top-most entry for each filename. If you wish to perform merging (similar to e.g. tags) or other operations that involve all the files for each filename from different resource/data packs, see the next section.
+Note that `SimpleJsonResourceReloadListener` goes through the resource packs top to bottom, and only retains the top-most entry for each filename. If you wish to perform merging (similar to e.g. [tags][tags]) or other operations that involve all the files for each filename from different resource/data packs, see the next section.
 
 ### `SimplePreparableReloadListener`
 
@@ -120,7 +120,7 @@ Note that `SimpleJsonResourceReloadListener` goes through the resource packs top
 
 `SimplePreparableReloadListener<T>` splits its reload into two distinct cycles: `prepare` and `apply`. First, `prepare` is called to collect the files and convert them into `T`s. Once **all** files have been collected, `apply` is called to do something with them - usually store them for later use.
 
-For a simple reference implementation of `SimplePreparableReloadListener` that loads from a single file, see `SplashManager`. For an implementation for merging JSON files, see `SoundManager#prepare`, `SoundManager#apply` and the related fields in `SoundManager`.
+For a simple reference implementation of `SimplePreparableReloadListener` that loads from a single file, see `SplashManager`. For an implementation for merging JSON files, see the merging of [`sounds.json`][soundsjson] in `SoundManager#prepare`, `SoundManager#apply` and the related fields in `SoundManager`.
 
 ### `ContextAwareReloadListener`
 
@@ -133,3 +133,11 @@ TODO
 ## `FileToIdConverter`
 
 TODO
+
+[events]: ../concepts/events.md
+[recipes]: server/recipes/index.md
+[resources]: index.md
+[sides]: ../concepts/sides.md
+[soundsjson]: client/sounds.md#soundsjson
+[tags]: server/tags.md
+[textures]: client/textures.md
--- a/docs/resources/reloadlisteners.md
+++ b/docs/resources/reloadlisteners.md
@@ -5,21 +5,28 @@ sidebar_position: 3
 
 In some situations, integrating with the [existing resource systems][resources] provided by Minecraft or NeoForge just isn't going to cut it. Instead, having your system load files by itself from a resource or data pack is more desirable. For this purpose, you can register a custom reload listener, implementing `PreparableReloadListener` or one of its subinterfaces/subclasses.
 
-The idea behind a reload listener is simple: When a resource pack or data pack reload happens, the listener is called upon to read its contents from the new set of resource or data packs, loaded into a global `ResourceManager` and usually reference-copied into other places for easier access. It will then keep the contents until the next reload, at which point the contents will be discarded and the cycle starts anew.
+The idea behind a reload listener is simple: When a resource pack or data pack reload happens, the listener is called upon to read its contents from the new set of resource or data packs. It will then keep the contents until the next reload, at which point the contents will be discarded and the cycle starts anew.
+
+:::info
+On the server [side][sides], the more robust [datapack registry][datapackregistries] or [data map][datamaps] systems should be preferred over a reload listener, if possible.
+:::
 
 ## Reloading
 
-Both resource pack and data pack reload function similar in principle and only differ in the associated [side][sides], and by extension the timing and the location files are loaded from.
+Both resource pack and data pack reload function similar in principle and only differ in the associated physical and logical [side][sides], and by extension the timing and the location files are loaded from.
 
-Client reload listeners load from resource packs (the `assets` folder). The first reload happens during startup on the physical client, and **never** on the physical server. Subsequent reloads are triggered when resource packs are changed in the Options menu, or by pressing F3+T.
+|                                              | Client Reload Listener                                                                                                             | Server Reload Listener                                                                                                |
+|----------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------|
+| **Loads From**                               | Resource packs (`assets` folder)                                                                                                   | Data packs (`data` folder)                                                                                            |
+| **First Reload**<br/>(Physical Client)       | - Startup                                                                                                                          | - Creating a new world<br/>("Preparing for world creation...")<br/>- Joining an existing world<br/>- Joining a server |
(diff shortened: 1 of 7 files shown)


docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
+---
+sidebar_position: 3
+---
+# Reload Listeners
+
+In some situations, integrating with the [existing resource systems][resources] provided by Minecraft or NeoForge just isn't going to cut it. Instead, having your system load files by itself from a resource or data pack is more desirable. For this purpose, you can register a custom reload listener, implementing `PreparableReloadListener` or one of its subinterfaces/subclasses.
```
ChampionAsh5357, line 6: While a solid header, I don't think this gets to the crux of what reload listeners are for. For example, there are many scenarios where it makes sense to use a datapack registry or a datamap instead of a server resource listener (client resource listeners are really the only entrypoint afaik). As such, I think we need to have a place that describes why I would use a resource listener compared to the other systems that exist.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+The idea behind a reload listener is simple: When a resource pack or data pack reload happens, the listener is called upon to read its contents from the new set of resource or data packs, loaded into a global `ResourceManager` and usually reference-copied into other places for easier access. It will then keep the contents until the next reload, at which point the contents will be discarded and the cycle starts anew.
```
ChampionAsh5357: It may be worth being more specific here by saying the pack repository since it could expand into the subpacks we briefly touch on the datagen docs.

ChampionAsh5357: So, it's less that its reference-copied for easier access, and more that the resource manager provides a method to look up and access a stream of the file contents. It typically does not store anything itself, more of a search feature on a file browser.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Reloading
+
+Both resource pack and data pack reload function similar in principle and only differ in the associated [side][sides], and by extension the timing and the location files are loaded from.
+
+Client reload listeners load from resource packs (the `assets` folder). The first reload happens during startup on the physical client, and **never** on the physical server. Subsequent reloads are triggered when resource packs are changed in the Options menu, or by pressing F3+T.
```
ChampionAsh5357: Also when downloading resource packs from a server.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+Server reload listeners load from data packs (the `data` folder). The first reload happens during world loading, which happens on world/server join on the physical client, and on startup on the physical server. Subsequent reloads are triggered using the `/reload` command, or by switching worlds/servers on the physical client.
```
ChampionAsh5357: First reload happens when opening the world creation screen, right before world loading.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+:::info
+The `/reload` command also triggers reloads of some other datapack-driven systems, such as [datapack registries][datapackregistries]; this is by design and cannot be circumvented. Datapack registries are reloaded before reload listeners, meaning you can use their values in your server-side reload listeners if needed.
```
ChampionAsh5357: Um, it does not. Datapack registries are loaded during initial world load and never again. Content that uses datapack registries do (e.g., tags, data components).

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+:::
+
+Reloading happens on multiple threads, as reload listeners are unrelated to one another. If you need to access another reload listener's values, consider delaying execution until the first use of your system after the reload has complete, or avoid access altogether.
```
ChampionAsh5357: You can share reload state across listeners, that is the purpose of `PreparableReloadListener#prepareSharedState`. Granted, only few things are shared, but this all is dependent on what you need. For example, may you instead use `TagsUpdatedEvent` to do the final computation because the state is not shared.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+// For server-side reload listeners
+@SubscribeEvent // on the game event bus
+public static void addServerReloadListeners(AddServerReloadListenersEvent event) {
+    event.addListener(MyReloadListener.ID, MyReloadListener.INSTANCE);
+}
+```
+
+:::danger
+Do not register the same reload listener on both sides! All of your file-driven systems should be designed for one side only, otherwise desyncs and similar issues will arise.
```
ChampionAsh5357, line 97: Technically, some reload listener data should be cleared on client log out to prevent state leakage (e.g., server reload listener syncs data to client).

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+    PreparableReloadListener --> ResourceManagerReloadListener;
+    PreparableReloadListener --> ContextAwareReloadListener --> SimplePreparableReloadListener --> SimpleJsonResourceReloadListener;
+    
+    class PreparableReloadListener,ResourceManagerReloadListener green;
+    class ContextAwareReloadListener,SimplePreparableReloadListener,SimpleJsonResourceReloadListener blue;
+```
+
+_Interfaces in green, abstract classes in blue._
+
+All reload listeners implement the `PreparableReloadListener` interface. However, in order to make things more understandable, we will first cover the `SimpleJsonResourceReloadListener`, which is used to load JSON files and implements most things for you already. Then, we will gradually go up the class hierarchy and explain what is done for you and what you can modify yourself in each level.
```
ChampionAsh5357: Not sure how much I agree with this. The `SimpleJsonResourceReloadListener` does cover a lot of use cases, but it obscures understanding quite a bit since it handles the basic async stuff for you.

I think it would be more useful to start from `PreparableReloadListener` and build up the intuition for what `reload` does and how the preparation and application phases work. In simple terms, create `SimplePreparableReloadListener` manually with a better description and then explaining the tools that already exist for you to use.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+        super(MyObject.CODEC, FileToIdConverter.json("mymod/my_listener"));
+    }
+
+    @Override
+    protected void apply(Map<Identifier, MyObject> map, ResourceManager resourceManager, ProfilerFiller profiler) {
+        // Clear out the old values.
+        values.clear();
+        // In the most simple implementation, blindly add all the values we get into the map.
+        // If needed, you can perform validations, duplicate removal or similar here.
+        values.putAll(map);
```
ChampionAsh5357: Sort of? This is where the preparation and application phases are a bit muddied. Technically, everything that can be done before the barrier should be done then, which includes validation and duplicate removal (which is already something the codec and the scanning method is doing). The only thing application should be doing is updating the prepared state for use.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+    public Map<Identifier, MyObject> getAll() {
+        return Collections.unmodifiableMap(values);
+    }
+}
+```
+
+And then simply access your values like so:
+
+```java
+MyObject myObject = MyReloadListener.INSTANCE.get(Identifier.fromNamespaceAndPath("mymod", "example"));
```
ChampionAsh5357, line 221: Of course, make sure to be careful to handle when the state is ill-defined, as can happen during reload.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+FileToIdConverter converter = FileToIdConverter.json("mymod/my_listener");
+```
+
+The above converter will convert paths as follows:
+
+- `assets/mymod/mymod/my_listener/example_1.json` -> `mymod:example_1`
+- `assets/mymod/mymod/my_listener/subfolder/example_2.json` -> `mymod:subfolder/example_2`
+- `assets/othermod/mymod/my_listener/example_3.json` -> `othermod:example_3`
+
+Besides `FileToIdConverter#json()`, there is an additional helper `FileToIdConverter#registry()` that accepts a [`ResourceKey<? extends Registry<?>>`][resourcekey] and converts the registry key to a namespace-and-path string, like the ones above.
```
ChampionAsh5357: Or you can just make an arbitrary one using the constructor.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+:::tip
+In order to avoid conflicts where two mods add a registry that is named the same, it is strongly recommended to prefix your reload listener's folder with a folder named after the mod id, as seen above with `mymod/my_listener`.
+:::
+
+### `SimplePreparableReloadListener`
+
+`SimplePreparableReloadListener<T>` sits a layer above `SimpleJsonResourceReloadListener<T>`, and is also usable for non-JSON files, as well as (JSON or non-JSON) files with the same name from different resource/data packs.
```
ChampionAsh5357: Wouldn't this be a layer below?

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+`SimplePreparableReloadListener<T>` splits its reload into two distinct cycles: `prepare` and `apply`. First, `prepare` is called to collect the files and convert them into `T`s. Once **all** files have been collected, `apply` is called to do something with them - usually store them for later use.
+
+For a simple reference implementation of `SimplePreparableReloadListener` that loads from a single file, see `SplashManager`. For an implementation for merging JSON files, see the merging of [`sounds.json`][soundsjson] in `SoundManager#prepare()`, `SoundManager#apply()` and the related fields in `SoundManager`. For an implementation of folder scanning, see `SimpleJsonResourceReloadListener#prepare()`.
```
ChampionAsh5357: `SplashManager` doesn't load a single file after Neo's modification. It loads all resources that match the splashes path.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+
+### `ResourceManagerReloadListener`
+
+Outside the class hierarchy described so far, `ResourceManagerReloadListener` is a utility interface that runs once the reload itself has completed, providing the fully-populated `ResourceManager` in its only method `#onResourceManagerReload()`. Classes implementing this interface mainly do post-reload cleanup work or build caches.
+
+### `PreparableReloadListener`
+
+Finally, `PreparableReloadListener` sits at the top of the hierarchy, and is the type accepted by the events above. It defines a method `#reload()` that returns a `CompletableFuture<Void>` and accepts four parameters:
+
+- `PreparableReloadListener.SharedState currentReload`: This holds the "global state" of the reload, most notably including the partially-initialized `ResourceManager`.
```
ChampionAsh5357: There is a lot more that goes on here than just a 'global state'.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,190 @@
...
+- `Executor taskExecutor`: The `Executor` for the bulk of the tasks. This executor runs on multiple threads.
+- `PreparableReloadListener.PreparationBarrier barrier`: A threading barrier object. When creating a `CompletableFuture` yourself, a call to `thenCompose(barrier::wait)` should be added to make sure all the `CompletableFuture`s have caught up. See vanilla uses of `barrier#wait()` for reference.
```
ChampionAsh5357: Okay, but that doesn't explain why I should call the barrier. Like, why does it matter if the futures have caught up (ignoring the fact that the game will deadlock if the barrier isn't queried)?

The answer lies in `SimplePreparableReloadListener`, where preparation is handled on the task executor (background executor), while application is handled on the reload executor (main thread).

ChampionAsh5357 (changes requested):

docs/resources/reloadlisteners.md
IchHabeHunger54: I opted to simplify this bit and not mention the `ResourceManager` at all, since it's not really relevant.

docs/resources/reloadlisteners.md
IchHabeHunger54: Moved to a separate section at the end of the article.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,313 @@
...
+sidebar_position: 3
+---
+# Reload Listeners
+
+In some situations, integrating with the [existing resource systems][resources] provided by Minecraft or NeoForge just isn't going to cut it. Instead, having your system load files by itself from a resource or data pack is more desirable. For this purpose, you can register a custom reload listener, implementing `PreparableReloadListener` or one of its subinterfaces/subclasses.
+
+The idea behind a reload listener is simple: When a resource pack or data pack reload happens, the listener is called upon to read its contents from the new set of resource or data packs. It will then keep the contents until the next reload, at which point the contents will be discarded and the cycle starts anew.
+
+:::warning
+On the server [side][sides], the more robust [datapack registry][datapackregistries] or [data map][datamaps] systems should be preferred over a reload listener, if possible.
```
XFactHD, line 11: I would mark this as info instead of warning because there's nothing wrong with using a custom reload listener instead. There is also nothing inherently more robust about the two alternatives, they are just existing systems that may or may not work for a given use case.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,313 @@
...
+| **Loads From**                               | Resource packs (`assets` folder)                                                                                                   | Data packs (`data` folder)                                                                                            |
+| **First Reload**<br/>(Physical Client)       | - Startup                                                                                                                          | - Creating a new world<br/>("Preparing for world creation...")<br/>- Joining an existing world<br/>- Joining a server |
+| **Subsequent Reloads**<br/>(Physical Client) | - Changing resource packs in the Options menu<br/>- Downloading a server's custom resource pack on server join<br/>- Pressing F3+T | - Leaving a world or server<br/>(unloads the listener)                                                                |
```
XFactHD, line 22: > - Leaving a world or server (unloads the listener)

This is wrong or at the very least incredibly confusing. Leaving a world or server does unload the data loaded during the last server-side resource reload but it does not touch the reload listeners at all. If a listener is neither kept in a field nor in Neo's "retained listeners" (more on that later) then it goes out of scope right after the reload completes, if it's retained then it goes out of scope with the data and if it's kept in a field then it stays around forever.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,313 @@
...
+public static void addClientReloadListeners(AddClientReloadListenersEvent event) {
+    event.addListener(MyReloadListener.ID, MyReloadListener.INSTANCE);
+}
+
+// For server-side reload listeners
+@SubscribeEvent // on the game event bus
+public static void addServerReloadListeners(AddServerReloadListenersEvent event) {
+    event.addListener(MyReloadListener.ID, MyReloadListener.INSTANCE);
+}
+```
```
XFactHD, line 94: Despite the additional warning below, this example should use different class names for server and client reload listeners to avoid confusion.

Server reload listeners should also ideally not be kept in a static field and instead use `AddServerReloadListenersEvent#addRetainedListener()` and later retrieve the listener from `ReloadableServerResources#getListener()`. Storing a server reload listener in a static field risks leaking loaded data, requiring manual cleanup, as well as reaching across sides.

docs/resources/reloadlisteners.md
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    // Then, use in reload() like so:
+    @Override
+    public CompletableFuture<Void> reload(SharedState currentReload, Executor taskExecutor, PreparationBarrier barrier, Executor reloadExecutor) {
+        MyPendingResources pending = currentReload.get(STATE_KEY);
+        // do the reload here, using `pending`
+    }
+}
+```
```
XFactHD, line 293: This example is incomplete (doesn't show how to "wait" for the other listener's data) and confusing (the state is provided and used by the same listener).

The shared object basically must be a "dangling" `CompletableFuture` (i.e. Neo's `AnimationLoader`) or map thereof (vanilla's `AtlasManager`). The dependent listener retrieves this future from the `SharedState` and incorporates it into its future chain to wait on it (see use of `AtlasManager.PENDING_STITCH` in `ModelManager#reload()`). The listener being depended on produces the data, retrieves the future from the `SharedState` and manually `complete()`s it (see `AnimationLoader#reload()`) which effectively "wakes up" the dependent listener.
The statement below this comment is incorrect (the `SharedState` is just a "dumb" map, it performs no synchronization whatsoever) and should be replaced with a rough explanation of this.

XFactHD (changes requested):

PR #373, opened 2026-09-05
fix doc using an invalid sound type for sounds.json

cylian914:
sound.json's sound entry only accept the type "file" and "event", the doc was using "sound" instead of "file"

------------------
Preview URL: https://pr-373.neoforged-docs-previews.pages.dev


sciwhiz12:
Related to https://github.com/neoforged/NeoForge/pull/3460. This bug went unnoticed since the introduction of the sound definition datagen because it didn't serialize the `type` when it equalled the default.

sciwhiz12 (dismissed):

ChampionAsh5357:
And the other versions?

PR #376, opened 2026-09-19
Display version dropdown on mobile

ChampionAsh5357:
Adds back in the version dropdown and color mode when viewing the page half width (mobile). 

------------------
Preview URL: https://pr-376.neoforged-docs-previews.pages.dev


sciwhiz12 (approved):
Seems to work for me. The version dropdown does cover the "Homepage" button, but that's a minor issue at best, and the version selector is more important.

